PR #2, opened 2019-03-07
Add outline of constraint framework

maxsmythe:
This is a rough outline of what I'm thinking for the framework.

Comments most welcome.

Most people are probably looking for the ConstraintTemplate outline:

constraint/pkg/apis/templates/v1alpha1/constrainttemplate_types.go


constraint/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Build the manager binary
```
tsandall: Nit: Is this file needed?

constraint/Gopkg.toml
```diff
@@ -0,0 +1,37 @@
+required = [
```
tsandall: Nit: Can we reduce this to the minimum required dependencies? It looks like this is copied from Gatekeeper. Would be nice to keep the library as lightweight as possible.

constraint/Makefile
```diff
@@ -0,0 +1,55 @@
+
```
tsandall: Nit: Same comments as on the go deps and Dockerfile apply here.

constraint/pkg/apis/templates/v1alpha1/constrainttemplate_types_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+
+	"github.com/onsi/gomega"
+	"golang.org/x/net/context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestStorageConstraintTemplate(t *testing.T) {
+	key := types.NamespacedName{
```
tsandall: Is this code generated or is there some desire to use gomega? I'd prefer to just use the Go standard library for testing and not rely on one of these frameworks.

constraint/Dockerfile
maxsmythe: Probably not. Let me go through and purge things a little more carefully

constraint/Gopkg.toml
maxsmythe: Required files are from kubebuilder. I can look and see what we may be able to drop as we are only using bits of kubebuilder here.

constraint/pkg/apis/templates/v1alpha1/constrainttemplate_types_test.go
maxsmythe: This is generated code.

maxsmythe:
Should have pruned extra libraries with this last amend

constraint/Gopkg.toml
maxsmythe: removed some extraneous libs

constraint/Makefile
maxsmythe: Cleaned

constraint/pkg/apis/templates/v1alpha1/constrainttemplate_types_test.go
tsandall: OK. If let's merge for now and then revisit. If we don't need to take a dependency on gomega, that would be preferable.

PR #5, opened 2019-03-14
Add client initialization logic

maxsmythe:

### DIFF
--- a/constraint/pkg/client/rego_helpers.go
+++ b/constraint/pkg/client/rego_helpers.go
@@ -98,3 +98,66 @@ func checkDataAccess(module *ast.Module) Errors {
 	}
 	return nil
 }
+
+// rule name -> arity
+type ruleArities map[string]int
+
+// requireRules makes sure the listed rules are specified with the required arity
+func requireRules(rego string, reqs ruleArities) error {
+	module, err := ast.ParseModule("foo", rego)
+	if err != nil {
+		return err
+	}
+
+	arities := make(ruleArities, len(module.Rules))
+	for _, rule := range module.Rules {
+		name := string(rule.Head.Name)
+		arity, err := getRuleArity(rule)
+		if err != nil {
+			return err
+		}
+		arities[name] = arity
+	}
+
+	var errs Errors
+	for name, arity := range reqs {
+		actual, ok := arities[name]
+		if !ok {
+			errs = append(errs, fmt.Errorf("Missing required rule: %s", name))
+			continue
+		}
+		if arity != actual {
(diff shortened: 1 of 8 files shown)


constraint/pkg/client/regolib/src.go
```diff
@@ -0,0 +1,38 @@
...
+package hooks.{{.Target}}
+
+audit[response] {
+	data.hooks.{{.Target}}.library.matching_reviews_and_constraints[[review, constraint]]
+	cnst_key := ["constraint"]
+	keys := [k | review[k]]
+	all_keys := array.concat(cnst_key, keys)
+	full_review := {k: v | v = audit_field_or_constraint(k, review, constraint); k = all_keys[_]}
+	review.constraint = constraint
+	data.templates.{{.Target}}[_].deny[response] with input as review
```
tsandall: How important is it to merge the constraint into the review? Alternatively we could supply the input document as...
```
input.review     # AdmissionReview object
input.constraint # Constraint object
```

This would also give us a bit of room in the future if we wanted to introduce other fields in the input document without worrying about conflicts.

constraint/pkg/client/rego_helpers.go
```diff
@@ -98,3 +98,66 @@ func checkDataAccess(module *ast.Module) Errors {
 	}
 	return nil
 }
+
+// rule name -> arity
+type ruleArities map[string]int
+
+// requireRules makes sure the listed rules are specified with the required arity
+func requireRules(rego string, reqs ruleArities) error {
+	module, err := ast.ParseModule("foo", rego)
```
tsandall: `"foo"` will show up in error messages. We'd be better off using something a bit more informative...some ideas:

"lib"
"hooks"
"built-in"

constraint/pkg/client/regolib/src.go
maxsmythe: Providing them separate sounds good to me. Per Morgante's comment on the design doc, we may want to choose a more generic name than "review" in that case, as other targets may not have the concept of a review request.

constraint/pkg/client/rego_helpers.go
maxsmythe: done

constraint/pkg/client/rego_helpers.go
```diff
@@ -98,3 +98,69 @@ func checkDataAccess(module *ast.Module) Errors {
 	}
 	return nil
 }
+
+// rule name -> arity
+type ruleArities map[string]int
+
+// requireRules makes sure the listed rules are specified with the required arity
+func requireRules(name, rego string, reqs ruleArities) error {
```
tsandall: In future we can make use of OPA's type inferencing here. This is fine for now.

PR #6, opened 2019-03-15
Add local driver

maxsmythe:

### DIFF
--- a/constraint/pkg/client/drivers/local/local.go
+++ b/constraint/pkg/client/drivers/local/local.go
@@ -38,7 +38,7 @@ func copyModules(modules map[string]*ast.Module, filter string) map[string]*ast.
 	return m
 }
 
-func (d *driver) PutRule(ctx context.Context, name string, src string) error {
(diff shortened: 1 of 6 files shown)


maxsmythe:
Almost done. Probably need to fix up the `regolib` package and write unit tests for client and integration tests for the remote client.

constraint/pkg/client/drivers/local/local.go
```diff
@@ -0,0 +1,208 @@
...
+	for k, v := range modules {
+		if filter != "" && k == filter {
+			continue
+		}
+		m[k] = v
+	}
+	return m
+}
+
+func (d *driver) PutRule(ctx context.Context, name string, src string) error {
```
tsandall: Nit: unrelated to the main content of this PR, but perhaps this should be called PutModule instead?

maxsmythe: I like it! Changed.

tsandall:
LGTM

PR #13, opened 2019-03-26
Hermetic test

maxsmythe:
This should allow us to run `make hermetic_test` to run all tests, including those requiring side-processes.


tsandall:
LGTM. I'll setup Travis to include a pre-merge check using the hermetic-test target.

PR #14, opened 2019-03-29
Use escaped paths for data

maxsmythe:

### DIFF
--- a/constraint/pkg/client/regolib/src.go
+++ b/constraint/pkg/client/regolib/src.go
@@ -47,11 +47,6 @@ inventory[inv] {
 }
 
 inventory[{}] {
-	not data.external
-}
-
-inventory[{}] {
-	data.external
 	not data.external["{{.Target}}"]
 }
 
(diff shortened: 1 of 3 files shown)


constraint/pkg/client/regolib/src.go
```diff
@@ -38,6 +40,20 @@ audit[response] {
 	}
 }
 
+# get_default(data, "external", {}) seems to cause this error:
+# "rego_type_error: undefined function data.hooks.<target>.get_default"
+inventory[inv] {
+	inv = data.external["{{.Target}}"]
+}
+
+inventory[{}] {
```
tsandall: Nit: This rule is not be needed given the one below which asserts that if if data.external[<TARGET>] is missing, inv should be`{}`.

maxsmythe: My thought here is "what if data.external exists, but data.external.<target> does not.

maxsmythe: Oh, wait, I see what you mean. Updated.

PR #17, opened 2019-05-01
Allow traces on a per-request basis

maxsmythe:


maxsmythe:
Good call. Found a bug and fixed it. Oops.

PR #20, opened 2019-06-05
Make text of ConstraintTemplate no target error clearer

maxsmythe:
This is to address https://github.com/open-policy-agent/gatekeeper/issues/123
on the Gatekeeper project.


constraint/pkg/client/crd_helpers.go
```diff
@@ -19,8 +19,10 @@ import (
 func validateTargets(templ *v1alpha1.ConstraintTemplate) error {
 	if len(templ.Spec.Targets) > 1 {
 		return errors.New("Multi-target templates are not currently supported")
+	} else if templ.Spec.Targets == nil {
```
ritazh: Shouldnt we check for nil before we check the length?

maxsmythe: Length is zero if the list is nil.

PR #26, opened 2019-06-28
handle namespaceselector and empty namespaces

ritazh:

Fixes https://github.com/open-policy-agent/gatekeeper/issues/153

### DIFF
--- a/constraint/pkg/client/regolib/src.go
+++ b/constraint/pkg/client/regolib/src.go
@@ -5,12 +5,12 @@ const (
 package hooks["{{.Target}}"]
 
 violation[response] {
-	data.hooks["{{.Target}}"].library.autoreject_review[[rejection, constraint]]
+	data.hooks["{{.Target}}"].library.autoreject_review[rejection]
 	review := get_default(input, "review", {})
 	response = {
 		"msg": get_default(rejection, "msg", ""),
 		"metadata": {"details": get_default(rejection, "details", {})},
-		"constraint": constraint,
+		"constraint": get_default(rejection, "constraint", {}),
 		"review": review,
 	}
 }
--- a/constraint/pkg/client/regolib/src.go
+++ b/constraint/pkg/client/regolib/src.go
@@ -4,12 +4,12 @@ const (
 	targetLibSrc = `
 package hooks["{{.Target}}"]
 
-deny[response] {
-	data.hooks["{{.Target}}"].library.autoreject_review[[violation, constraint]]
+violation[response] {
+	data.hooks["{{.Target}}"].library.autoreject_review[[rejection, constraint]]
 	review := get_default(input, "review", {})
 	response = {
-		"msg": get_default(violation, "msg", ""),
-		"metadata": {"details": get_default(violation, "details", {})},
+		"msg": get_default(rejection, "msg", ""),
+		"metadata": {"details": get_default(rejection, "details", {})},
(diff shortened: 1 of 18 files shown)


constraint/pkg/client/regolib/src.go
```diff
@@ -4,10 +4,24 @@ const (
 	targetLibSrc = `
 package hooks["{{.Target}}"]
 
+# Skip constraint evaluation for a given target if constraint has namespaceSelector but Namespace is not cached
```
maxsmythe: k8s-specific code should not be in the framework. The framework should provide a way for the target-handler to auto-reject and put the NS selector auto-rejection in the targethandler

ritazh: How about adding a new rule similar to `matching_constraints[constraint]` in `Library()` ? For example, the new rule could be called ` autoreject_constraints[constraint, review]` and implemented by the target handler.

maxsmythe: Per our conv. on Friday, that SGTM

constraint/pkg/client/regolib/src.go
```diff
@@ -4,8 +4,14 @@ const (
...
 package hooks["{{.Target}}"]
 
+# Auto reject reviews when a constraint has namespaceSelector but Namespace is not cached
+deny[response] {
+	data.hooks["{{.Target}}"].library.autoreject_review[response]
+}
+
 # Finds all violations for a given target
 deny[response] {
+	not data.hooks["{{.Target}}"].library.autoreject_review[response]
```
maxsmythe: this `not` line should be dropped. We should keep reporting violations even if autoreject is also true as the intent is to return as much actionable feedback (violations) as possible.

constraint/pkg/client/regolib/src.go
```diff
@@ -4,8 +4,14 @@ const (
 	targetLibSrc = `
 package hooks["{{.Target}}"]
 
+# Auto reject reviews when a constraint has namespaceSelector but Namespace is not cached
+deny[response] {
```
maxsmythe: The format for `response` should be handled by this Rego package, as the CF relies on the interface defined by response to parse/return violations.

maxsmythe:
Somehow didn't see that `regolib/src.go` was edited. Dunno why. Sorry :/

constraint/pkg/client/regolib/src.go
ritazh: so maybe `autoreject_review` should just return the msg?

constraint/pkg/client/regolib/src.go
```diff
@@ -4,6 +4,17 @@ const (
 	targetLibSrc = `
 package hooks["{{.Target}}"]
 
+violation[response] {
+	data.hooks["{{.Target}}"].library.autoreject_review[[rejection, constraint]]
```
maxsmythe: `constraint` should be a property of `rejection`. This way handlers that reject for reasons that have nothing to do with the constraint don't need to worry about creating a dummy constraint.

constraint/pkg/client/regolib/src.go
```diff
@@ -4,6 +4,17 @@ const (
...
+	review := get_default(input, "review", {})
+	response = {
```
maxsmythe: We should test that these results actually get passed on to the client. Unit tests & e2e tests.

PR #27, opened 2019-07-02
fix CT name validation

ritazh:

Fixes https://github.com/open-policy-agent/gatekeeper/issues/156


constraint/pkg/client/client.go
```diff
@@ -223,6 +223,9 @@ func (c *client) CreateCRD(ctx context.Context, templ *v1alpha1.ConstraintTempla
 	if templ.ObjectMeta.Name != templ.Spec.CRD.Spec.Names.Plural {
```
maxsmythe: I am split on requiring the plural name == the lower-cased kind (which should equal the singular name).

On the one hand it's an odd requirement. On the other hand it plays well with our convention of appending a version to the name.

One way versioning might change is by giving users control over the group, allowing them to store the version there.

My impulse is to keep what's here b/c tighter requirements can always be loosened. Thoughts?

ritazh: > keep what's here b/c tighter requirements can always be loosened

Agreed.

PR #28, opened 2019-07-02
Rename deny rule to violation

maxsmythe:


tsandall:
@maxsmythe I assume you'll coordinate with the Forseti folks on this?

maxsmythe:
Ya, I'll give them a heads up, though they are not currently using CF directly

PR #33, opened 2019-07-09
Check to make sure templates have a violation rule defined

### LINKED ISSUE #32: Validate rule signature for Constraint Templates
We should check that constraint templates implement a `violation` rule and raise an error if they don't.

maxsmythe:
Fixes #32

### DIFF
--- a/constraint/pkg/client/client_test.go
+++ b/constraint/pkg/client/client_test.go
@@ -294,6 +294,23 @@ func TestRemoveData(t *testing.T) {
 func TestAddTemplate(t *testing.T) {
 	badRegoTempl := createTemplate(name("fakes"), crdNames("Fake", "fakes"), targets("h1"))
 	badRegoTempl.Spec.Targets[0].Rego = "asd{"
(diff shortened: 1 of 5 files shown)


constraint/pkg/client/client_test.go
```diff
@@ -330,6 +347,18 @@ func TestAddTemplate(t *testing.T) {
...
+			Handler:       &badHandler{Name: "h1", HasLib: true},
+			Template:      badArityTempl,
+			ErrorExpected: true,
+		},
+		{
+			Name:          "Missing Rule",
+			Handler:       &badHandler{Name: "h1", HasLib: true},
+			Template:      missingRuleTempl,
+			ErrorExpected: true,
+		},
```
ritazh: Should we also have a positive test for case`violation[r]`?

maxsmythe: Taken care of in `Good Template`

ritazh:
LGTM

PR #35, opened 2019-07-09
Create a CRD to define MutationTemplates

cmccart750:
Duplicates the ConstraintTemplates CRD and renamed
the kinds to define MutationTemplate(s), templates
that will be used to define a series of Mutations
(Similar to ConstraintTemplates & Constraints)
for a mutating Gatekeeper.

Tested by applying `crds/templates_v1alpha1_mutationtemplate.yaml`
and `samples/templates_v1alpha1_mutationtemplate.yaml` on a GKE cluster,
then searching for `test-mute` in `kubectl get mutationtemplates`.

### DIFF
--- a/constraint/config/crds/templates_v1alpha1_mutationtemplate.yaml
+++ b/constraint/config/crds/templates_v1alpha1_mutationtemplate.yaml
@@ -1,0 +1,64 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: mutationtemplates.templates.gatekeeper.sh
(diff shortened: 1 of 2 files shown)


constraint/config/crds/templates_v1alpha1_mutationtemplate.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
maxsmythe: This code should be autogened via `make manifests`. Was this done by hand? This goes hand-in-hand w/kubebuilder

adrianludwin:
This was by hand, yes. Where are the original files? Is it the Go files?

maxsmythe:
Correct, the manifests are generated from the Go files under `pkg/api`

PR #36, opened 2019-07-10
Introduce MutationTemplate Go type

cmccart750:
Defines MutationTemplates, MutationTemplateSpecs, etc. by copying type
definitions and replacing "Constraint" with "Mutation". Also re-ran `go
generate` to ensure these types had DeepCopy methods (the only thing
blocking it from compiling)


constraint/pkg/apis/templates/v1alpha1/constrainttemplate_types.go
```diff
@@ -29,6 +29,11 @@ type ConstraintTemplateSpec struct {
 	Targets []Target `json:"targets,omitempty"`
 }
 
+type MutationTemplateSpec struct {
```
maxsmythe: mutationtemplate type should be created via `kubebuilder create`, which would put all of this in a separate directory.

adrianludwin:
What about the structs that are in common between them? Or should we not
have any?

maxsmythe:
Common structs can be imported and included.

It's important to use kubebuilder so that the expected directory structure is preserved for autogenerated code.

PR #38, opened 2019-07-12
Kubebuilder files for MutationTemplates

cmccart750:
Kubebuilder-generated code to define MutationTemplates - a CRD .yaml file and some Go structures that share the same internals as ConstraintTemplates. Only one line was changed by hand - setting spec.scope to Cluster in `templates_v1alpha1_mutationtemplate.yaml` to match its ConstraintTemplate counterpart. Everything should compile fine.


constraint/config/samples/templates_v1alpha1_mutationtemplate.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: templates.gatekeeper.sh/v1alpha1
+kind: MutationTemplate
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: mutationtemplate-sample
+spec:
+  # Add fields here
```
adrianludwin: Does this part need to be fixed?

cmccart750: Do you think we should add more fields? It's currently the exact same as the constrainttemplate-sample

adrianludwin: In that case, no :)

PR #39, opened 2019-07-13
Add v1beta1 versions of Constraints and ConstraintTemplates

maxsmythe:
Note that this also required moving to a model where we use a common
core representation of the template resource and implement a method for
conversion between each version.

Another issue was that the version of controller-gen that ships
with Kubebuilder v1 doesn't handle multi-version CRDs, which I've
addressed by patching the versions in via kustomize.


constraint/pkg/apis/templates/v1beta1/constrainttemplate_types_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func TestTypeConversion(t *testing.T) {
+	scheme := runtime.NewScheme()
+	AddToSchemes.AddToScheme(scheme)
+
+	versioned := &ConstraintTemplate{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConstraintTemplate",
+			APIVersion: "templates.gatekeeper.sh/v1alpha1",
```
ritazh: nit: s/v1alpha1/v1beta1/

constraint/pkg/client/crd_helpers.go
```diff
@@ -60,49 +71,62 @@ func newCRDHelper() *crdHelper {
...
+					Storage: false,
+					Served:  true,
+				},
+			},
 		},
 	}
-	h.scheme.Default(crd)
-	crd.ObjectMeta.Name = fmt.Sprintf("%s.%s", crd.Spec.Names.Plural, constraintGroup)
-	return crd
+	// Defaulting functions only exist for v1beta1
```
ritazh: can you provide more information for this?

constraint/pkg/client/crd_helpers.go
```diff
@@ -118,8 +142,8 @@ func (h *crdHelper) validateCR(cr *unstructured.Unstructured, crd *apiextensions
 	if cr.GroupVersionKind().Group != constraintGroup {
 		return fmt.Errorf("Wrong group for constraint %s. Have %s, want %s", cr.GetName(), cr.GroupVersionKind().Group, constraintGroup)
 	}
-	if cr.GroupVersionKind().Version != crd.Spec.Version {
-		return fmt.Errorf("Wrong version for constraint %s. Have %s, want %s", cr.GetName(), cr.GroupVersionKind().Version, crd.Spec.Version)
+	if !supportedVersions[cr.GroupVersionKind().Version] {
```
ritazh: Shouldn't this check against the supported versions of `constraints.gatekeeper.sh`not `templates.gatekeeper.sh`?

ritazh:
Left few comments, but otherwise LGTM.

constraint/pkg/apis/templates/v1beta1/constrainttemplate_types_test.go
maxsmythe: oops, fixed.

constraint/pkg/client/crd_helpers.go
maxsmythe: I'm assuming these will be 1:1, but that's not necessarily guaranteed.

constraint/pkg/client/crd_helpers.go
maxsmythe: The kubernetes scheme is documented here:

https://godoc.org/k8s.io/apimachinery/pkg/runtime#Scheme

I've tried to use the unversioned, internal representation of the CRD types in an effort to make the API a bit more stable.

The CRD library has chosen to only register defaulting functions (those that set defaults on a type) for the versioned verision of the CRD resource. This function takes an unversioned schema, which corresponds to the unversioned CRD.

This means I have two options:

1. Convert the schema to a versioned JSONSchema, implement a versioned CRD, apply defaults, convert to unversioned.
2. Create an unversioned CRD, convert to versioned CRD, apply defaults, convert to unversioned CRD

I expect choice (2) will require less change if/when CRD v2 comes around.

constraint/pkg/client/crd_helpers.go
ritazh: This is probably ok for now, but we might need to come back to this later.

maxsmythe: ack, LMK if you'd much rather prefer I add some constants somewhere. I don't really have a preference personally.

PR #40, opened 2019-07-13
Only pass parameters to ConstraintTemplate Rego source

maxsmythe:

### DIFF
--- a/constraint/pkg/client/regolib/src.go
+++ b/constraint/pkg/client/regolib/src.go
@@ -21,7 +21,7 @@ violation[response] {
 	review := get_default(input, "review", {})
 	inp := {
 		"review": review,
-		"parameters": constraint.spec.parameters
+		"parameters": get_default(get_default(constraint, "spec", {}), "parameters", {}),
 	}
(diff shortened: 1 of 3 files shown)


constraint/pkg/client/regolib/src.go
```diff
@@ -21,7 +21,7 @@ violation[response] {
 	review := get_default(input, "review", {})
 	inp := {
 		"review": review,
-		"constraint": constraint
+		"parameters": constraint.spec.parameters
```
maxsmythe: this should use `get_default`

constraint/pkg/client/regolib/src.go
```diff
@@ -39,7 +39,7 @@ audit[response] {
 	data.hooks["{{.Target}}"].library.matching_reviews_and_constraints[[review, constraint]]
 	inp := {
 		"review": review,
-		"constraint": constraint,
+		"parameters": constraint.spec.parameters,
```
maxsmythe: this should use `get_default`

PR #42, opened 2019-07-17
Remove Client interface and replace with concrete struct

adrianludwin:
Updated to follow guidance at https://github.com/golang/go/wiki/CodeReviewComments#interfaces.

Tested by importing into Gatekeeper's vendor directory, changing the
references to Client to pointers, and rebuilding and running all tests.

Signed-off-by: Adrian Ludwin <adrian.ludwin@gmail.com>


adrianludwin:
NB: This will break all clients (see https://github.com/open-policy-agent/gatekeeper/pull/187 for how to unbreak Gatekeeper). I am assuming there are no other significant clients of CF other than GK right now, but if that's incorrect, they'll have to make similar changes as well when importing the latest CF.

maxsmythe:
LGTM. You are correct in that GK is the only current client of CF I'm aware of.

I'm looking to start putting semvers on this repo soon. Likely as part of the move of GK to beta.

PR #44, opened 2019-07-22
Allow for the creation of MutationCRDs alongside Constraint CRDs

cmccart750:
Refactored OPA client code and introduced a few Mutation variants of Client functions to allow them to accept MutationTemplates and then build appropriate Mutation CRD Go objects. Tested mainly with `go test` at the package level, and then later copying diffs into a new Gatekeeper image and running that on a GKE cluster.

### DIFF
--- a/constraint/config/crds/templates_v1alpha1_mutationtemplate.yaml
+++ b/constraint/config/crds/templates_v1alpha1_mutationtemplate.yaml
@@ -17,12 +17,12 @@ spec:
         apiVersion:
           description: 'APIVersion defines the versioned schema of this representation
             of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+            internal value, and may reject unrecognized values. More info: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md'
           type: string
         kind:
           description: 'Kind is a string value representing the REST resource this
-            object represents. Servers may infer this from the endpoint the client
-            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          object represents.Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md'
           type: string
         metadata:
           type: object
--- a/constraint/pkg/apis/templates/v1alpha1/mutationtemplate_types_test.go
+++ b/constraint/pkg/apis/templates/v1alpha1/mutationtemplate_types_test.go
@@ -26,13 +26,11 @@ import (
 
 func TestStorageMutationTemplate(t *testing.T) {
 	key := types.NamespacedName{
-		Name:      "foo",
-		Namespace: "default",
+		Name: "foo",
 	}
 	created := &MutationTemplate{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "foo",
-			Namespace: "default",
+			Name: "foo",
 		}}
 	g := gomega.NewGomegaWithT(t)
 
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -23,7 +23,7 @@ type k8sGroup string
 
 const (
 	constraintGroup k8sGroup = "constraints.gatekeeper.sh"
-	mutationGroup k8sGroup = "mutations.gatekeeper.sh"
+	mutationGroup   k8sGroup = "mutations.gatekeeper.sh"
 )
 
 type Client interface {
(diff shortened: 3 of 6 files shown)


cmccart750:
The conflict is a change to the tests for MT types - changing an expected namespace from "default" to "" because I remember we ran into a similar bug elsewhere in some auto-generated test code?

adrianludwin:
We'll resolve the conflicts tomorrow, but this should be enough to start a
review.

constraint/pkg/client/client.go
```diff
@@ -214,19 +220,43 @@ func createTemplatePath(target, name string) string {
...
+	if err := c.validateGenericCRD(nm, specCRD, targets); err != nil {
 		return nil, err
 	}
-	if templ.ObjectMeta.Name == "" {
-		return nil, errors.New("Template has no name")
+	return c.createGenericCRD(nm, specCRD, targets, mutationGroup)
+}
+
+func (c *client) validateGenericCRD(objMetaName string, specCRD *v1alpha1.CRD, targets []v1alpha1.Target) error {
+	crdSpecKind := specCRD.Spec.Names.Kind
```
maxsmythe: this should just be kind

constraint/pkg/client/client.go
```diff
@@ -214,19 +220,43 @@ func createTemplatePath(target, name string) string {
...
+	targets := templ.Spec.Targets
+	if err := c.validateGenericCRD(nm, specCRD, targets); err != nil {
 		return nil, err
 	}
-	if templ.ObjectMeta.Name == "" {
-		return nil, errors.New("Template has no name")
+	return c.createGenericCRD(nm, specCRD, targets, mutationGroup)
+}
+
+func (c *client) validateGenericCRD(objMetaName string, specCRD *v1alpha1.CRD, targets []v1alpha1.Target) error {
```
maxsmythe: specCRD should just be crd

maxsmythe: objmetaname would just be name

constraint/pkg/client/client.go
```diff
@@ -214,19 +220,43 @@ func createTemplatePath(target, name string) string {
 
 // CreateCRD creates a CRD from template
 func (c *client) CreateCRD(ctx context.Context, templ *v1alpha1.ConstraintTemplate) (*apiextensionsv1beta1.CustomResourceDefinition, error) {
-	if err := validateTargets(templ); err != nil {
+	nm := templ.ObjectMeta.Name
```
maxsmythe: nm -> name

constraint/pkg/client/client.go
```diff
@@ -214,19 +220,43 @@ func createTemplatePath(target, name string) string {
...
+	specCRD := &templ.Spec.CRD
+	targets := templ.Spec.Targets
+	if err := c.validateGenericCRD(nm, specCRD, targets); err != nil {
+		return nil, err
+	}
+	return c.createGenericCRD(nm, specCRD, targets, constraintGroup)
+}
+
+func (c *client) CreateMutationCRD(ctx context.Context, templ *v1alpha1.MutationTemplate) (*apiextensionsv1beta1.CustomResourceDefinition, error) {
```
maxsmythe: It's probably better to have CreateCRD() take an interface{} typed variable and do type switching

constraint/pkg/client/client.go
maxsmythe: We can fold the `Generic...` functions back into one if we do type switching.

constraint/pkg/client/client.go
```diff
@@ -214,19 +220,43 @@ func createTemplatePath(target, name string) string {
...
 	}
-	if templ.ObjectMeta.Name != strings.ToLower(templ.Spec.CRD.Spec.Names.Kind) {
-		return nil, fmt.Errorf("Template's name %s is not equal to the lowercase of CRD's Kind: %s", templ.ObjectMeta.Name, strings.ToLower(templ.Spec.CRD.Spec.Names.Kind))
+	if objMetaName != strings.ToLower(crdSpecKind){
+		return fmt.Errorf("Template's name %s is not equal to the lowercase of CRD's Kind: %s", objMetaName, strings.ToLower(crdSpecKind))
 	}
+	return nil
+}
 
+func (c *client) createGenericCRD(objMetaName string, specCRD *v1alpha1.CRD, targets []v1alpha1.Target, group k8sGroup) (*apiextensionsv1beta1.CustomResourceDefinition, error){
```
maxsmythe: `name` + `crd` for variable names.

constraint/config/crds/templates_v1alpha1_mutationtemplate.yaml
```diff
@@ -0,0 +1,87 @@
...
+    kind: MutationTemplate
+    plural: mutationtemplates
+  scope: Cluster
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
```
pmcalpine: This doc has moved to:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
(here and in the description below)

constraint/pkg/apis/templates/v1alpha1/mutationtemplate_types_test.go
```diff
@@ -0,0 +1,57 @@
...
+	created := &MutationTemplate{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Namespace: "",
+		}}
+	g := gomega.NewGomegaWithT(t)
+
+	// Test Create
+	fetched := &MutationTemplate{}
+	g.Expect(c.Create(context.TODO(), created)).NotTo(gomega.HaveOccurred())
```
pmcalpine: I see that the existing tests use context.TODO(), but a single context.Background() instantiated at the start of this test would be consistent with what's recommended by the golang docs:
https://golang.org/pkg/context/#Background

constraint/pkg/client/client.go
```diff
@@ -455,7 +493,13 @@ func (c *client) validateConstraint(constraint *unstructured.Unstructured, lock
 // ValidateConstraint returns an error if the constraint is not recognized or does not conform to
 // the registered CRD for that constraint.
 func (c *client) ValidateConstraint(ctx context.Context, constraint *unstructured.Unstructured) error {
-	return c.validateConstraint(constraint, true)
+	return c.validateConstraint(constraint, true, constraintGroup)
+}
+
+// ValidateMutation returns an error if the constraint is not recognized or does not conform to
+// the registered CRD for that constraint.
+func (c *client) ValidateMutation(ctx context.Context, constraint *unstructured.Unstructured) error {
```
pmcalpine: Do you plan to make ValidateMutation do something different than ValidateConstraint in the future?

constraint/pkg/apis/templates/v1alpha1/mutationtemplate_types_test.go
adrianludwin: I believe this is autogenerated code from kubebuilder

constraint/pkg/client/client.go
adrianludwin: Cam, FYI: https://tour.golang.org/methods/16

constraint/pkg/client/client.go
adrianludwin: We had trouble here because `crd` is also the name of the thingy created by CreateCRD internally. So we needed to differentiate between the CRD from the spec we were passing in, and the one that got passed back. `specCRD` seemed like a decent compromise and we used it in multiple places.

adrianludwin:
> Note that these changes will be affected pretty severely by the changes recently checked into master for beta.

Agreed, we'll worry about that later.

constraint/pkg/client/client.go
pmcalpine: Discussed offline.

constraint/pkg/apis/templates/v1alpha1/mutationtemplate_types_test.go
maxsmythe: +1 to "it's autogenned"

constraint/pkg/client/crd_helpers.go
```diff
@@ -16,25 +16,25 @@ import (
...
-	if len(templ.Spec.Targets) > 1 {
+func validateTargets(targets []v1alpha1.Target) error {
+	if len(targets) > 1 {
 		return errors.New("Multi-target templates are not currently supported")
-	} else if templ.Spec.Targets == nil {
-		return errors.New(`Field "targets" not specified in ConstraintTemplate spec`)
-	} else if len(templ.Spec.Targets) == 0 {
-		return errors.New("No targets specified. ConstraintTemplate must specify one target")
+	} else if targets == nil {
+		return errors.New(`Field "targets" not specified in Template spec`)
```
maxsmythe: we can lower-case template here as it's no longer referring to a specific kind

constraint/pkg/client/client.go
maxsmythe: Per a conversation with Cam, we can consolidate this into one CreateCRD function later. This would be a requirement for merging into master.

PR #45, opened 2019-08-01
Add support for enforcementaction

ritazh:

### DIFF
--- a/constraint/pkg/client/crd_helpers.go
+++ b/constraint/pkg/client/crd_helpers.go
@@ -38,7 +38,8 @@ func validateTargets(templ *templates.ConstraintTemplate) error {
 // to form the schema of the actual constraint resource
 func (h *crdHelper) createSchema(templ *templates.ConstraintTemplate, target MatchSchemaProvider) (*apiextensions.JSONSchemaProps, error) {
 	props := map[string]apiextensions.JSONSchemaProps{
-		"match": target.MatchSchema(),
+		"match":             target.MatchSchema(),
+		"enforcementAction": apiextensions.JSONSchemaProps{Type: "string"},
 	}
 	if templ.Spec.CRD.Spec.Validation != nil && templ.Spec.CRD.Spec.Validation.OpenAPIV3Schema != nil {
 		internalSchema := &apiextensions.JSONSchemaProps{}
@@ -47,7 +48,6 @@ func (h *crdHelper) createSchema(templ *templates.ConstraintTemplate, target Mat
 		}
 		props["parameters"] = *internalSchema
 	}
-	props["enforcementaction"] = apiextensions.JSONSchemaProps{Type: "string"}
 	schema := &apiextensions.JSONSchemaProps{
 		Properties: map[string]apiextensions.JSONSchemaProps{
 			"spec": apiextensions.JSONSchemaProps{
--- a/constraint/pkg/client/crd_helpers.go
+++ b/constraint/pkg/client/crd_helpers.go
@@ -47,6 +47,7 @@ func (h *crdHelper) createSchema(templ *templates.ConstraintTemplate, target Mat
 		}
 		props["parameters"] = *internalSchema
 	}
+	props["enforcementaction"] = apiextensions.JSONSchemaProps{Type: "string"}
 	schema := &apiextensions.JSONSchemaProps{
 		Properties: map[string]apiextensions.JSONSchemaProps{
 			"spec": apiextensions.JSONSchemaProps{
--- a/constraint/pkg/client/crd_helpers_test.go
+++ b/constraint/pkg/client/crd_helpers_test.go
@@ -103,7 +103,7 @@ func tProp(t string) apiextensions.JSONSchemaProps {
 }
 
 func expectedSchema(pm propMap) *apiextensions.JSONSchemaProps {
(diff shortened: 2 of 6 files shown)


constraint/pkg/client/crd_helpers.go
```diff
@@ -47,6 +47,7 @@ func (h *crdHelper) createSchema(templ *templates.ConstraintTemplate, target Mat
 		}
 		props["parameters"] = *internalSchema
 	}
+	props["enforcementaction"] = apiextensions.JSONSchemaProps{Type: "string"}
```
maxsmythe: `enforcementAction` (camel case)

maxsmythe: Also, this can just be added to the `props` assignment on line 40

constraint/pkg/client/regolib/src.go
```diff
@@ -7,11 +7,15 @@ package hooks["{{.Target}}"]
...
+	constraint := get_default(rejection, "constraint", {})
+	spec := get_default(constraint, "spec", {})
+	enforcementaction := get_default(spec, "enforcementaction", "deny")
 	response = {
 		"msg": get_default(rejection, "msg", ""),
 		"metadata": {"details": get_default(rejection, "details", {})},
-		"constraint": get_default(rejection, "constraint", {}),
+		"constraint": constraint,
 		"review": review,
+		"enforcementaction": enforcementaction,
```
maxsmythe: camel case

constraint/pkg/types/validation.go
```diff
@@ -22,6 +22,9 @@ type Result struct {
 
 	// The violating Resource, filled out by the Target
 	Resource interface{}
+
+	// The enforcement action of the constraint
+	EnforcementAction string `json:"enforcementaction,omitempty"`
```
maxsmythe: camel case

constraint/pkg/client/regolib/src.go
```diff
@@ -7,11 +7,15 @@ package hooks["{{.Target}}"]
 violation[response] {
 	data.hooks["{{.Target}}"].library.autoreject_review[rejection]
 	review := get_default(input, "review", {})
+	constraint := get_default(rejection, "constraint", {})
+	spec := get_default(constraint, "spec", {})
+	enforcementaction := get_default(spec, "enforcementaction", "deny")
```
maxsmythe: "deny" or "DENY"?

constraint/pkg/client/crd_helpers_test.go
```diff
@@ -103,6 +103,7 @@ func tProp(t string) apiextensions.JSONSchemaProps {
 }
 
 func expectedSchema(pm propMap) *apiextensions.JSONSchemaProps {
+	pm["enforcementaction"] = apiextensions.JSONSchemaProps{Type: "string"}
```
maxsmythe: Is this necessary? This is meant to allow building of arbitrary schema, and I don't see "enforcementAction" tested anywhere?

constraint/pkg/client/regolib/src.go
```diff
@@ -43,11 +50,14 @@ audit[response] {
 	}
 	inventory[inv]
 	data.templates["{{.Target}}"][constraint.kind].violation[r] with input as inp with data.inventory as inv
+	spec := get_default(constraint, "spec", {})
+	enforcementaction := get_default(spec, "enforcementaction", "deny")
```
maxsmythe: Could we test:

1) That enforcementAction non-default gets passed through
2) that the default enforcmentaction is deny

in unit tests?

constraint/pkg/client/regolib/src.go
ritazh: so are we requiring all `enforcementAction` to be cap?

constraint/pkg/client/crd_helpers_test.go
ritazh: this was necessary as the tests were failing because `createSchema` now adds `enforcementAction` to `spec` by default, but `expectedSchema` calls `prop` which does not include `enforcementAction`.
https://github.com/open-policy-agent/frameworks/blob/master/constraint/pkg/client/crd_helpers_test.go#L93-L98

constraint/pkg/client/regolib/src.go
maxsmythe: ya, it's golang and k8s api convention to use camel case

maxsmythe: My thought is that traditionally keywords like this have been cap, so I thought it might make sense to continue the tradition. (ex ip tables)

Curious if anyone else has any opinion.

constraint/pkg/client/crd_helpers_test.go
maxsmythe: ack, SGTM

constraint/pkg/client/regolib/src.go
ritazh: done

maxsmythe:
LGTM. An earlier commit says WIP is there anything more you had in mind, or is this the whole change?

ritazh:
WIP is on the PR in Gatekeeper: https://github.com/open-policy-agent/gatekeeper/pull/202 will remove that label after this PR is merged.

PR #47, opened 2019-08-14
Add option to restrict allowed data fields

maxsmythe:

### DIFF
--- a/constraint/pkg/client/backend.go
+++ b/constraint/pkg/client/backend.go
@@ -41,9 +41,14 @@ func (b *Backend) NewClient(opts ...ClientOpt) (Client, error) {
 	if b.hasClient {
 		return nil, errors.New("Currently only one client per backend is supported")
 	}
+	var fields []string
+	for k := range validDataFields {
+		fields = append(fields, k)
+	}
 	c := &client{
-		backend:     b,
-		constraints: make(map[string]*constraintEntry),
+		backend:           b,
+		constraints:       make(map[string]*constraintEntry),
+		allowedDataFields: fields,
 	}
 	var errs Errors
 	for _, opt := range opts {
@@ -54,6 +59,7 @@ func (b *Backend) NewClient(opts ...ClientOpt) (Client, error) {
 	if len(errs) > 0 {
 		return nil, errs
 	}
+	c.rConformer = newRegoConformer(c.allowedDataFields)
 	if len(c.targets) == 0 {
 		return nil, errors.New("No targets registered. Please register a target via client.Targets()")
 	}
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -95,6 +95,20 @@ func Targets(ts ...TargetHandler) ClientOpt {
 	}
 }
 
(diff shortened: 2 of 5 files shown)


constraint/pkg/client/client.go
```diff
@@ -95,6 +95,20 @@ func Targets(ts ...TargetHandler) ClientOpt {
...
 }
 
+// AllowedDataFields sets the fields under `data` that Rego in ConstraintTemplates
+// can access. If unset, all fields can be accessed. Only fields recognized by
+// the system can be enabled.
+func AllowedDataFields(fields ...string) ClientOpt {
+	return func(c *client) error {
+		var errs Errors
+		c.allowedDataFields = fields
+		if len(errs) > 0 {
```
ritazh: where does this `errs` get updated?

constraint/pkg/client/client_test.go
```diff
@@ -606,3 +606,66 @@ func TestRemoveConstraint(t *testing.T) {
...
+			Handler:       &badHandler{Name: "h1", HasLib: true},
+			Template:      createTemplate(name("fakes"), crdNames("Fakes"), targets("h1")),
+			ErrorExpected: false,
+		},
+		{
+			Name:          "Inventory Used",
+			Handler:       &badHandler{Name: "h1", HasLib: true},
+			Template:      inventoryTempl,
+			ErrorExpected: true,
+		},
```
ritazh: Should we add a test case where `Inventory` used is allowed?

constraint/pkg/client/client.go
maxsmythe: oops, this should go

constraint/pkg/client/client_test.go
maxsmythe: Good idea, added.

constraint/pkg/client/backend.go
```diff
@@ -54,6 +59,7 @@ func (b *Backend) NewClient(opts ...ClientOpt) (Client, error) {
 	if len(errs) > 0 {
 		return nil, errs
```
ritazh: Isn't this returning an array but expecting an `error`?

constraint/pkg/client/client_test.go
```diff
@@ -606,3 +606,78 @@ func TestRemoveConstraint(t *testing.T) {
...
+	for _, tt := range tc {
+		t.Run(tt.Name, func(t *testing.T) {
+			d := local.New()
+			b, err := NewBackend(Driver(d))
+			if err != nil {
+				t.Fatalf("Could not create backend: %s", err)
+			}
+			f := AllowedDataFields()
+			if tt.InvAllowed {
+				f = AllowedDataFields("inventory")
```
ritazh: can we add another test for if `f = AllowedDataFields("notallowed")`? and make sure NewClient returns an empty c.allowedDataFields

constraint/pkg/client/backend.go
maxsmythe: The array obeys the Error interface

constraint/pkg/client/client_test.go
maxsmythe: done

PR #51, opened 2019-09-05
Do not allow empty Rego source strings on ConstraintTemplates

maxsmythe:
It causes a panic

### DIFF
--- a/constraint/pkg/client/rego_helpers.go
+++ b/constraint/pkg/client/rego_helpers.go
@@ -143,6 +143,9 @@ type ruleArities map[string]int
 
 // requireRules makes sure the listed rules are specified with the required arity
 func requireRules(name, rego string, reqs ruleArities) error {
+	if rego == "" {
+		return errors.New("Rego source code is empty")
+	}
 	module, err := ast.ParseModule(name, rego)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


maxsmythe:
This will partially address https://github.com/open-policy-agent/gatekeeper/issues/238 when merged into GK.

I'd still like the version info so that I can use line numbers to fix the nil pointer dereference closer to the source (if possible).

dolv:
quay.io/open-policy-agent/gatekeeper:v3.0.4-beta.0

maxsmythe:
Thanks! Found the root of the nil-dereference and removed it :)

constraint/pkg/client/rego_helpers.go
```diff
@@ -143,10 +149,16 @@ type ruleArities map[string]int
...
 func requireRules(name, rego string, reqs ruleArities) error {
+	if rego == "" {
+		return errors.New("Rego source code is empty")
+	}
 	module, err := ast.ParseModule(name, rego)
 	if err != nil {
 		return err
 	}
+	if module == nil {
+		return "", errors.New("Module could not be parsed")
```
ritazh: oh this is giving us an error. should just return error, no string

maxsmythe: oops, that's what I get for not using an IDE...

fixed, will push if tests pass.

PR #54, opened 2019-09-14
First pass at integrating rego rewriter into constraint framework.

briantkennedy:
* Note: unit tests are not yet updated in anticipation of feedback on
implementation choices.

Signed-off-by: Brian Thomas Kennedy <briankennedy@google.com>


briantkennedy:
@maxsmythe PTAL when you have a chance

constraint/pkg/client/client.go
```diff
@@ -236,59 +270,121 @@ func (c *Client) CreateCRD(ctx context.Context, templ *templates.ConstraintTempl
...
+	defer func() {
+		for _, path := range rollbackPaths {
+			if _, err := c.backend.driver.DeleteModule(ctx, path); err != nil {
+				glog.Warningf("encountered error while rolling back module %s", path)
+			}
+		}
+	}()
+
+	for _, mod := range artifacts.modules {
+		if err := c.backend.driver.PutModule(ctx, mod.path, mod.rego); err != nil {
```
maxsmythe: Currently PutModule triggers a full recompile each time. This is problematic here because:

1. The compile might fail if there are missing library functions the current module relies on
2. Uses CPU unnecessarily

I think the right thing to do here is to add a `PutModules()` function that is atomic. We can do this with the local driver no problem. @tsandall has mentioned that they were interested in adding this to OPA's rest API. In the interim the remote driver's `PutModules()` could just be a loop around `PutModule()`. This would make it impossible to use the remote driver with Kubernetes until the atomic update is fixed (not a blocker as K8s doesn't currently have the option to run remote OPA anyway.). I believe Torin and I had discussed this limitation and are fine with it (but cc-ing to raise awareness).

constraint/pkg/client/client.go
```diff
@@ -236,59 +270,121 @@ func (c *Client) CreateCRD(ctx context.Context, templ *templates.ConstraintTempl
...
+			return resp, err
```
maxsmythe: Another problem with this AddModule loop is that currently the Gatekeeper update handler for constraint templates just calls `AddTemplate()` to replace the original code. That behavior is incompatible with this update as-written because it could lead to dangling library entries in the case where an updated template is removing a library (there is no tombstone for deletion).

One possible suggestion: if we have `PutModules()` (suggested in above comment) take a "root" argument, where everything under that root is replaced atomically, we can keep the old behavior. I'm open to other ideas, but that's what occurs to me at the moment.

One performance-enhancing item we can add to the local driver is to check for changes to the template source (if an overwrite) and only re-run the compiler if there is a delta (optional enhancement to add to this PR if we go with the above change).

constraint/pkg/client/client.go
```diff
@@ -236,59 +270,121 @@ func (c *Client) CreateCRD(ctx context.Context, templ *templates.ConstraintTempl
...
 
-	return crd, nil
+	var mods []constraintTemplateModule
+	if err := sources.ForEachModule(func(m *regorewriter.Module) error {
+		content, err := m.Content()
+		if err != nil {
+			return err
+		}
+		mods = append(mods, constraintTemplateModule{
+			path: m.Module.Package.Path.String(),
```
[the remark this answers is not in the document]
briantkennedy: I think adding PutModules will make this go away.

constraint/pkg/client/client.go
briantkennedy: re: point 1 - I'll add PutModules/DeleteModules (more detail in other comment).
re: point 2 - this is the same as the status quo, so I think we are conflating things if we want to address this here.

constraint/pkg/client/client.go
briantkennedy: So it sounds like I need to add PutModules/DeleteModules, correct?  How do the following signatures look?
```
PutModules(ctx context.Context, namePrefix string, sources []string) error
DeleteModules(ctx context.Context, namePrefix string) (count int, err error)
```

Using a single key for many modules seems like it would be the right interface, I'll look into that.

As for the performance aspect, I think I'd like focus this change on correctness since we could always optimize later on.

constraint/pkg/client/client_test.go
```diff
@@ -719,34 +721,14 @@ func TestAllowedDataFieldsIntersection(t *testing.T) {
...
 				expected[v] = true
 			}
-			if !reflect.DeepEqual(c.rConformer.allowedDataFields, expected) {
-				t.Errorf("c.rConformer.allowedDataFields = %v; want %v", c.rConformer.allowedDataFields, expected)
-			}
 		})
 	}
 }
-
-func TestAllowedDataFieldsNotSpecified(t *testing.T) {
```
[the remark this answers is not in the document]
briantkennedy: done

constraint/pkg/client/crd_helpers.go
```diff
@@ -22,12 +22,11 @@ var supportedVersions = map[string]bool{
 	v1beta1.SchemeGroupVersion.Version:  true,
 }
 
-// validateTargets ensures that the targets field has the appropriate values
+// validateTargets ensures that the targets field has the appropriate values.  At the moment we only
+// allow exactly one target.
 func validateTargets(templ *templates.ConstraintTemplate) error {
 	if len(templ.Spec.Targets) > 1 {
 		return errors.New("Multi-target templates are not currently supported")
-	} else if templ.Spec.Targets == nil {
```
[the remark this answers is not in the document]
briantkennedy: golang discourages comparing a slice to nil and prescribes using len() == 0.  Added this back.

constraint/pkg/client/rego_helpers.go
```diff
@@ -59,23 +59,39 @@ func (rc *regoConformer) ensureRegoConformance(kind, path, rego string) (string,
 	return module.String(), nil
 }
 
+func parseModule(path, rego string) (*ast.Module, error) {
```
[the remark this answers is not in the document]
briantkennedy: removed

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: Potentially, though all "files" still need a unique name. The unique name forms the URI for the OPA rest interface, and is used as the map key for the internal client.

constraint/pkg/client/client.go
maxsmythe: +1 to both responses

constraint/pkg/client/client.go
maxsmythe: Re: performance SGTM, though in the local one, we should aim to run compiles at most once per call to `PutModules()`, as the inefficiency of doing otherwise is quite high and avoiding the multiple calls should be easy. Remote is out-of-scope, as that would require changes to the OPA REST interface.

My only concern with the prefix design is we are assuming that the consumers of PutModules() and PutModule() are aware of this clobbering-by-prefix potential. I'm fine with the risk as I think the CF should completely encapsulate these methods, but if there is an easy alternative it's worth pursuing.

constraint/pkg/client/client.go
maxsmythe: re: point 2: Sorry, I was thinking that point 2 was about avoiding recompilation on update. Current status quo is to recompile at most once per AddTemplate(), we should try to preserve that, as recompile can be expensive, and I'm assuming many libraries per constraint.

constraint/pkg/client/client_test.go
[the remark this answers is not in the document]
maxsmythe: Did the tests get re-added? Didn't see it in the commit.

constraint/pkg/client/client.go
briantkennedy: I'm not sure I see where it's currently doing the recompile check, is that inside of PutModule()?

constraint/pkg/client/client.go
briantkennedy: Added PutModules / DeleteModules function in local.go, PTAL and let me know if that impl is good then I can add some unit tests and integrate here.

constraint/pkg/client/drivers/local/local.go
```diff
@@ -62,64 +82,146 @@ func copyModules(modules map[string]*ast.Module, filter string) map[string]*ast.
...
+func (d *driver) checkModuleName(name string) error {
+	if strings.HasPrefix(name, moduleSetPrefix) {
+		return errors.Errorf("Single modules not allowed to use name prefix %s", moduleSetPrefix)
+	}
+	return nil
+}
+
+func (d *driver) checkModuleSetName(name string) error {
+	if strings.HasPrefix(name, moduleSetSep) {
+		return errors.Errorf("Module sets not allowed to use name prefix %s", moduleSetPrefix)
```
maxsmythe: s/moduleSetPrefix/moduleSetSep

constraint/pkg/client/client.go
```diff
@@ -157,34 +158,71 @@ func createTemplatePath(target, name string) string {
...
+
+// validateTargets handles validating the targets section of the CT.
+func (c *Client) validateTargets(templ *templates.ConstraintTemplate) (*templates.Target, TargetHandler, error) {
 	if err := validateTargets(templ); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
+
+	if len(templ.Spec.Targets) != 1 {
+		panic("consistency error, expected exactly 1 item in targets")
```
maxsmythe: We probably don't want to panic, as that would cause one bad template to make the controller/webhook crash loop.

constraint/pkg/client/rego_helpers.go
```diff
@@ -15,151 +14,35 @@ var (
...
-func (errs Errors) Error() string {
-	s := make([]string, len(errs))
-	for _, e := range errs {
-		s = append(s, e.Error())
-	}
-	return strings.Join(s, "\n")
-}
-
-// checkDataAccess makes sure that data is only referenced in terms of valid subfields
-func (rc *regoConformer) checkDataAccess(module *ast.Module) Errors {
```
maxsmythe: IIRC your Rego rewriter reimplements this. This is correct right?

If so, no issue

constraint/pkg/client/drivers/remote/remote.go
```diff
@@ -90,6 +95,11 @@ func (d *driver) DeleteModule(ctx context.Context, name string) (bool, error) {
 	return err == nil, err
 }
 
+// DeleteModules implements drivers.Driver
```
maxsmythe: The not implemented errors are causing the tests of the remote driver to fail. We can disable those tests for now, I think.

constraint/pkg/client/client.go
briantkennedy: now returns an error

constraint/pkg/client/rego_helpers.go
briantkennedy: Yeah, it checks usage of 'data' to ensure that it will only reference libs or allowed externs.

constraint/pkg/client/client.go
maxsmythe: I should have closed this out earlier.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: List works with the PutModules() implementation.

constraint/pkg/client/client.go
maxsmythe: I believe this is addressed via `PutModules()`

maxsmythe:
It looks like we're just missing tests for `AddModules()` / `RemoveModules()` and multi-source constraint templates.

maxsmythe:
Looks like we're missing the multi-source constraint template tests in `client/e2e_tests.go`.

After that, LGTM

briantkennedy:
Added lib testing to testcases to e2e_tests.go

PR #55, opened 2019-09-25
Add more docs to TargetHandler

briantkennedy:
Signed-off-by: Brian Thomas Kennedy <briankennedy@google.com>

### DIFF
--- a/constraint/pkg/client/targethandler.go
+++ b/constraint/pkg/client/targethandler.go
@@ -1,0 +1,99 @@
+package client
+
+import (
+	"text/template"
+
+	"github.com/open-policy-agent/frameworks/constraint/pkg/types"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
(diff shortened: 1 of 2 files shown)


constraint/pkg/client/targethandler.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Library returns the pieces of Rego code required to stitch together constraint
+	// evaluation for the target.
+	//
+	// Templated Parameters:
+	//	`{{.ConstraintsRoot}}`: All constraints are found here and are placed under
+	//		data["{{.ConstraintsRoot}}"][KIND][NAME] where $KIND is the Kind indicated in the
+	//		CT .spec.crd.spec.names.kind field and NAME is the .metadata.name of the
+	//		Constraint resource.
+	//	`{{.DataRoot}}`: The data root, use data["{{.DataRoot}}"] instead of
```
maxsmythe: Correct. Side note: the data is actually cached in a target-specific path, so `data.inventory` only exists at template execution time.

(no action needed)

PR #63, opened 2020-01-04
Test gomod

### LINKED ISSUE #62: Broken downstream dependencies due to replace directive in go.mod
I'm getting some broken build dependencies in downstream projects after rolling forward to 1307ba72bce38ee3cf44f94def1bbc41eb4ffa90.

```
go build ./...
# k8s.io/kubectl/pkg/scheme
../../../golang/work/pkg/mod/k8s.io/kubectl@v0.0.0-20190926002329-2b775f287352/pkg/scheme/scheme.go:38:9: undefined: unstructured.NewJSONFallbackEncoder
# k8s.io/client-go/rest
../../../golang/work/pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/rest/request.go:598:31: not enough arguments in call to watch.NewStreamWatcher
	have (*versioned.Decoder)
	want (watch.Decoder, watch.Reporter)
```

After inspection, `k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible` is being replaced by `k8s.io/client-go v0.0.0-20191016110837-54936ba21026` in go.mod.

```
replace (
    k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.0.0-20191016113439-b64f2075a530
    k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20191004115701-31ade1b30762
    k8s.io/client-go => k8s.io/client-go v0.0.0-20191016110837-54936ba21026
)
```

The versions in the replace list seem to be required as a dependencies, so they should be in the `require` list rather than `replace` list otherwise dependant modules will have broken dependencies at incompatible versions.

[From the docs](https://github.com/golang/go/wiki/Modules#gomod): _exclude and replace directives only operate on the current (“main”) module. exclude and replace directives in modules other than the main module are ignored when building the main module._

maxsmythe:
Fixes #62


maxsmythe:
@tsandall this change eliminates the need for `replace` directives in the gomod files of any project that uses this library.

PR #64, opened 2020-01-06
Fix logging in regorewriter.

briantkennedy:
The logging was pretty much all at Info which caused way too much log
spam. This change moves the log statements to V levels 1 and 2.


briantkennedy:
@maxsmythe ptal when you've got some time, not urgent.

maxsmythe:
Actually, need DCO

PR #69, opened 2020-01-15
Issue 67

josh-ferrell:
Just ShortNames was added for now, adding configurable categories later.


constraint/pkg/apis/templates/v1alpha1/constrainttemplate_types_test.go
```diff
@@ -76,8 +76,9 @@ func TestTypeConversion(t *testing.T) {
 		Spec: ConstraintTemplateSpec{
 			CRD: CRD{
 				Spec: CRDSpec{
-					Names: Names{
+					Names: Names
```
maxsmythe: Needs bracket

PR #71, opened 2020-01-17
Expose state of constraints and templates to the user

maxsmythe:


constraint/pkg/client/client.go
```diff
@@ -227,12 +227,34 @@ func (c *Client) createTemplateArtifacts(templ *templates.ConstraintTemplate) (*
...
+}
+
+// createTemplateArtifacts will validate the CT, create the CRD for the CT's constraints, then
+// validate and rewrite the rego sources specified in the CT.
+func (c *Client) createTemplateArtifacts(templ *templates.ConstraintTemplate) (*ctArtifacts, error) {
+	artifacts, err := c.createBasicTemplateArtifacts(templ)
+	if err != nil {
+		return nil, err
+	}
+	targetSpec, targetHandler, err := c.validateTargets(templ)
```
ritazh: nit: If we also save `targetSpec` in `basicCTArtifacts` then we can skip this call and use `artifacts.targetSpec`

maxsmythe: added

maxsmythe:
LMK if approved, then I can work on the logging PRs in GK

PR #82, opened 2020-02-28
When a template is removed, remove all its constraints

### LINKED ISSUE #80: Cascading delete for constraints when removing parent template
In the interest of supporting asynchronous removal of Constraint Templates and Constraints from OPA, the framework client (`Client.RemoveTemplate`) should support cascading removal of constraints from OPA when removing their parent template.

maxsmythe:
Fixes #80 


maxsmythe:
@shomron could not add you as a reviewer, but this is for your PR in GK

ritazh:
lint issue

constraint/pkg/client/client.go
```diff
@@ -490,7 +525,7 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 		resp.Handled[target] = true
 	}
 	if len(errMap) == 0 {
-		c.constraints[subPath] = constraint.DeepCopy()
+		c.constraints[constraint.GroupVersionKind().GroupKind()][subPath] = constraint.DeepCopy()
```
ritazh: should we check for empty GroupVersionKind for completeness

maxsmythe: getTemplateEntry() would have raised an error

maxsmythe:
re: lint issue: weird, I didn't even touch that file. Fixed.

constraint/pkg/client/client.go
```diff
@@ -358,12 +363,15 @@ func (c *Client) AddTemplate(ctx context.Context, templ *templates.ConstraintTem
 		CRD:      artifacts.crd,
 		Targets:  []string{artifacts.targetHandler.GetName()},
 	}
+	if _, ok := c.constraints[artifacts.gk]; !ok {
```
shomron: Silly question - what happens if multiple ConstraintTemplates are defined with the same constraint GroupKind? Is that prevented anywhere?
Do we need to include the ConstraintTemplate name in the key for this map?

constraint/pkg/client/client.go
```diff
@@ -379,6 +387,17 @@ func (c *Client) RemoveTemplate(ctx context.Context, templ *templates.Constraint
 		return resp, err
 	}
 
+	for _, cstr := range c.constraints[artifacts.gk] {
+		if r, err := c.removeConstraintNoLock(ctx, cstr); err != nil {
```
shomron: Should we remove individual entries upon success in case we exit early?

constraint/pkg/client/client.go
```diff
@@ -539,7 +578,7 @@ func (c *Client) getConstraintNoLock(ctx context.Context, constraint *unstructur
 		return nil, err
 	}
 
-	cstr, ok := c.constraints[subPath]
+	cstr, ok := c.constraints[constraint.GroupVersionKind().GroupKind()][subPath]
```
shomron: Yay to Go supporting lookup on nil maps! :)

constraint/pkg/client/client_test.go
```diff
@@ -487,6 +488,48 @@ func TestRemoveTemplate(t *testing.T) {
...
+	}
+	cst2 := newConstraint("CascadingDelete", "cascadingdelete2", nil, nil)
+	if _, err = c.AddConstraint(context.Background(), cst2); err != nil {
+		t.Error("could not add second constraint")
+	}
+
+	if _, err = c.RemoveTemplate(context.Background(), template); err != nil {
+		t.Error("could not remove template")
+	}
+	if len(c.constraints) != 0 {
```
shomron: Worth adding another template to show that cascading delete of one template leaves those intact?

constraint/pkg/client/client.go
shomron: FWIW `getTemplateEntry()` only looks at the `Kind` so there's still the possibility of shenanigans.

maxsmythe: Technically an empty group is valid, though at that point it's not a Gatekeeper resource unless K8s changes.

I can add validation of the group to getTemplateEntry()

constraint/pkg/client/client.go
maxsmythe: Constraint Templates must be named after the constraint kind, which forces them to be singletons.

constraint/pkg/client/client.go
maxsmythe: Oh yes! Assigning however... :p

constraint/pkg/client/client_test.go
maxsmythe: +1

constraint/pkg/client/client.go
maxsmythe: Remove individual entries? Like from the cache? If so, removeConstraintNoLock does that (it's basically delete constraint but without the lock). Cleaning up the map is just for memory leaks.

shomron: Ok you mean [here](https://github.com/open-policy-agent/frameworks/blob/a0f9a6b1e4e7f1fa76999e1b0ac9d78fdbeea75f/constraint/pkg/client/client.go#L568)? Missed that earlier, thanks!

constraint/pkg/client/client.go
maxsmythe: added

constraint/pkg/client/client.go
maxsmythe: Yep!

shomron:
🎉

PR #95, opened 2020-09-21
Add a new category "constraints" to the generated CRDs.

fedepaol:
A user may use kubectl get contraint or get constraints to retrieve all the
crds generated by templates.

Fixes the second item discussed in https://github.com/open-policy-agent/gatekeeper/issues/677 (plus, this will need to be consumed)


maxsmythe (approved):
LGTM, thanks for doing this!

PR #97, opened 2020-10-19
deps: update opa to v0.24.0

srenatus:
Naively ran `go get -u github.com/open-policy-agent/opa` and `go mod vendor`.

I wonder if it would make sense here to have some automation do this instead? Similar to what happens in https://github.com/open-policy-agent/opa-envoy-plugin...


srenatus:
Hmm something is wrong here on travis. Any ideas? Have a fallen into some well-known pit?

ritazh:
> I wonder if it would make sense here to have some automation do this instead

@srenatus That's a great idea! Added this [issue](https://github.com/open-policy-agent/frameworks/issues/98) to track

constraint/go.mod
```diff
@@ -3,15 +3,26 @@ module github.com/open-policy-agent/frameworks/constraint
 go 1.13
 
 require (
+	github.com/OneOfOne/xxhash v1.2.8 // indirect
```
ritazh, line 6: Should this PR just focus on updating `github.com/open-policy-agent/opa v0.24.0`?

srenatus: I might have done something wrong, but that was my intention. I believe this is pulled in as a transitive dependency of OPA, but I can double check in the morning.

srenatus: https://github.com/open-policy-agent/opa/blob/v0.24.0/go.mod#L6 👈 yeah I think this is correct. It's a dependency of OPA, so upgrading OPA would pull this in.

srenatus:
Looks like travis is happy now 🎉

maxsmythe (approved):
LGTM

maxsmythe:
@ritazh LGTY?

ritazh (approved):
lgtm

PR #106, opened 2021-01-20
Fix addtoscheme

sozercan:
- fixes AddToScheme so autogenerated conversions are used
- removes unnecessary conversion between JSONSchemaProps and template OpenAPIV3Schema

### DIFF
--- a/constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
+++ b/constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
@@ -15,11 +15,7 @@ limitations under the License.
 
 package apis
 
-import (
-	"github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1"
-)
-
 func init() {
 	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
-	AddToSchemes = append(AddToSchemes, v1alpha1.AddToSchemes...)
+	AddToSchemes = append(AddToSchemes, AddToSchemes...)
 }
--- a/constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
+++ b/constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
@@ -15,7 +15,9 @@ limitations under the License.
 
 package apis
 
+import "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1"
+
 func init() {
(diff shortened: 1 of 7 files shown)


constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
```diff
@@ -15,11 +15,7 @@ limitations under the License.
 
 package apis
 
-import (
-	"github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1"
-)
-
 func init() {
 	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
-	AddToSchemes = append(AddToSchemes, v1alpha1.AddToSchemes...)
```
maxsmythe, line 24: I'm a bit confused here. Why are we appending a list to itself?

Should this not read as:

`AddToSchemes = append(AddToSchemes, v1alpha1.AddToScheme)`

?

constraint/pkg/apis/addtoscheme_templates_v1beta1.go
```diff
@@ -15,11 +15,7 @@ limitations under the License.
 
 package apis
 
-import (
-	"github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1"
-)
-
 func init() {
 	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
-	AddToSchemes = append(AddToSchemes, v1beta1.AddToSchemes...)
```
maxsmythe, line 24: same question here WRT self-appending here.

constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
```diff
@@ -15,11 +15,7 @@ limitations under the License.
 
 package apis
 
-import (
-	"github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1"
```
maxsmythe, line 19: dropping the import seems suspicious to me. If we aren't importing v1beta1, how are we adding its scheme?

maxsmythe (changes requested):
Thanks for doing this! One question about the AddToSchemes aggregations.

constraint/pkg/apis/addtoscheme_templates_v1alpha1.go
sozercan: @maxsmythe thanks, it should be fixed now

maxsmythe (approved):
LGTM

PR #108, opened 2021-02-05
add additional columns to constraints

sozercan:

adds total violations and enforcement action as part of `kubectl get` on constraints

image


maxsmythe (approved):
This is awesome!

It occurs to me that we need to be intentional in terms of how we roll this out to GK

If we simply vendor this code in and push, the old GK pods and the new ones will fight over the CRD for the duration that both Pods are simultaneously in existence.

This is probably relatively harmless as this period should be relatively brief (the time it takes a Deployment to roll out the change), and I think the fights are non-destructive, adding QPS to the API server, but this should be tested.

As an alternative, we can push out a GK update that ignores diffs caused by `AdditionalPrinterColumns`, then wait for that go go out before vendoring this in.

It would take longer to roll out, but would avoid the contention.

ritazh (approved):

ritazh:
> adding QPS to the API server, but this should be tested.

+1

sozercan:
Any thoughts on how would we test the QPS to the API server and how do we ignore diffs in GK?

maxsmythe:
Testing would involve manually halting an upgrade mid-deployment. I think this can be simulated by just creating 3 Gatekeeper Pods (without a deployment) and seeing what happens to the traffic. Or having two deployments and slowly cross-fading between the two.

As to how we can ignore diffs... it's a 2 release dance:

**Release 1:**
Update the constraint template controller so that it ignores the part of the diff that will be changing. `go-cmp` may make this easier. Release Gatekeeper.

**Release 2:**
Remove the parts of the code that ignore the diff, use the latest version of the constraint framework. Release Gatekeeper.

That way, release two will add the new feature where missing and release 1 wont try to revert it.

sozercan:
What happens if users go directly to release 2 and skip release 1?

maxsmythe:
There will be potential write contention between the two versions. The impact would have to be tested, but I believe it's just increased QPS to the API server.

A more dangerous possibility would be CRs getting deleted because the CRD changed, but I don't think K8s does that.

This model is basically a database refactoring. It rhymes with Martin Fowler's [Transition phase](https://martinfowler.com/articles/evodb.html#TransitionPhase).

willbeason:
DCO is incorrect. I have signed off all commits.

willbeason:
My guess is that DCO doesn't like that I added a commit on my windows machine, which uses "@gmail.com" unlike my corp machine.

ritazh:
@willbeason I see this got merged yesterday. Has this https://github.com/open-policy-agent/frameworks/pull/108#pullrequestreview-584792632 been addressed before merging into gk?

maxsmythe:
@ritazh 

Impact: no customer data loss, some controller fights mid-rollout as different controllers try to write the "correct" version of the CT. Controller fight goes away once the rollout completes. Symptoms of controller fight: increased requests to the API server, increasing resource version/generation.

Controller fights could be mitigated by moving CT -> CRD creation to a singleton controller (place it on the audit pod, like the status controller is).

PR #111, opened 2021-04-23
Upgrade: controller-gen v0.5.0, conversion-gen v0.20.2

julianKatz:
frameworks/constraint had previously been reliant on an outdated version
of the controller-gen library for generating CRDs with embedded
JSONSchemas.  The older version of the library wasn't built to detect
a nested JSONSchema, and thus would not refuse it.

With the requirement of v1 CRDs around the corner in k8s 1.22,
framework/constraint requires the newer controller-gen v0.5.0.  This
library version can output v1 CRD.

The conversion logic was also in need of a library upgrade, with
conversion-gen v0.21.0 now generating the necessary files.

This PR _does not_ output a v1 CRD for Constraint Template.  That will
be left for a future PR.  This PR brings the two binaries up-to-date.

Note: I attempted to dockerize these two binaries.  Controller-gen
worked fine, but for some reason conversion-gen couldn't write to the
filesystem.  The output looked fine, but no dice.  Rather than figure
out why, I decided to punt on the dockeriziation effort.


constraint/config/crd_patch.yaml
```diff
@@ -3,13 +3,4 @@ kind: CustomResourceDefinition
...
-    status: {}
-  versions:
-    - name: v1beta1
-      served: true
-      storage: true
-    - name: v1alpha1
-      served: true
-      storage: false
-
+  version: v1beta1
```
julianKatz, line 6: Would love to get rid of this completely, but I'm not sure how.

@maxsmythe you know the kubebuilder annotations pretty well... do you know how to change it?

That said, all the v1beta1 CRD examples in [this kubernetes documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/) don't even include the `version:` field, so maybe we can just remove it.

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
```diff
@@ -42,6 +46,8 @@ spec:
                           type: array
                       type: object
                     validation:
+                      properties:
```
maxsmythe, line 49: type: object (this can be done with a controller-gen tag)

constraint/config/crd_patch.yaml
maxsmythe: Yeah, the version field just needs to equal the first item in the `versions` list, if present.

From the looks of it, this can be removed.

maxsmythe (approved):
Basically LGTM with two nits.

@sozercan Can you see if this alone is enough to unblock your upgrading controller-runtime in GK? I'm skeptical because I'm not seeing new methods in the autogenned code.

If not, maybe controller-gen takes its cues from the version of controller-runtime used. And we'll need to upgrade that dependency and regen.

constraint/config/crd_patch.yaml
maxsmythe: It looks like leaving it in is actually breaking the unit tests

sozercan:
@maxsmythe I think it'll need to regenerate the conversions for it to work with new apimachinery. Sounds like when we regenerate with new version, it should remove s.Convert() or have it accept 2 args.

```
#17 59.04 # github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1
#17 59.04 vendor/github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1/zz_generated.conversion.go:438:22: too many arguments in call to s.Convert
#17 59.04 	have (*v1beta1.JSONSchemaProps, *apiextensions.JSONSchemaProps, number)
#17 59.04 	want (interface {}, interface {})
#17 59.04 vendor/github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1/zz_generated.conversion.go:457:22: too many arguments in call to s.Convert
#17 59.04 	have (*apiextensions.JSONSchemaProps, *v1beta1.JSONSchemaProps, number)
#17 59.04 	want (interface {}, interface {})
#17 59.07 # github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1
#17 59.07 vendor/github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1/zz_generated.conversion.go:438:22: too many arguments in call to s.Convert
#17 59.07 	have (*v1beta1.JSONSchemaProps, *apiextensions.JSONSchemaProps, number)
#17 59.07 	want (interface {}, interface {})
#17 59.07 vendor/github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1/zz_generated.conversion.go:457:22: too many arguments in call to s.Convert
#17 59.07 	have (*apiextensions.JSONSchemaProps, *v1beta1.JSONSchemaProps, number)
#17 59.07 	want (interface {}, interface {})
```

maxsmythe:
@sozercan WDYT of committing this as-is (as this change at least gets us to a workable version of controller-gen), then we can play around with figuring out whatever kind of poking we need to do to update the conversion function signatures?

maxsmythe:
@sozercan @julianKatz It looks like controller-gen doesn't integrate conversion-gen, which is what is used to generate the conversion functions. We'll need to keep building/applying that one manually (if possible):

https://pkg.go.dev/k8s.io/code-generator/cmd/conversion-gen

Luckily conversion-gen appears to be versioned in lockstep with K8s, so we should be able to upgrade it along with controller-runtime.

maxsmythe:
Also, because we can just move dependencies in lockstep now, we wont need to run conversion-gen as it's own, self-contained gomodule/docker container.

sozercan:
can we re-add the conversion-gen parts back to makefile (and whatever else is missing)? So we can only focus on upgrading controller-gen in this PR and follow up on conversion-gen later.

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
julianKatz: Done

constraint/config/crd_patch.yaml
julianKatz: Removed it using kustomize

constraint/pkg/apis/templates/v1beta1/zz_generated.conversion.go
```diff
@@ -434,10 +434,8 @@ func autoConvert_v1beta1_Validation_To_templates_Validation(in *Validation, out
 	if in.OpenAPIV3Schema != nil {
 		in, out := &in.OpenAPIV3Schema, &out.OpenAPIV3Schema
 		*out = new(apiextensions.JSONSchemaProps)
-		// TODO: Inefficient conversion - can we improve it?
-		if err := s.Convert(*in, *out, 0); err != nil {
-			return err
-		}
+		// FIXME: Provide conversion function to convert apiextensionsv1beta1.JSONSchemaProps to apiextensions.JSONSchemaProps
+		compileErrorOnMissingConversion()
```
sozercan, line 438: this is an issue generating these conversion. not sure why this happens since these conversions exists.

example:
```
func autoConvert_v1beta1_Validation_To_templates_Validation(in *Validation, out *templates.Validation, s conversion.Scope) error {
        if in.OpenAPIV3Schema != nil {
                in, out := &in.OpenAPIV3Schema, &out.OpenAPIV3Schema
                *out = new(apiextensions.JSONSchemaProps)
                if err := v1.Convert_v1_JSONSchemaProps_To_apiextensions_JSONSchemaProps(*in, *out, s); err != nil {
                        return err
                }
        } else {
                out.OpenAPIV3Schema = nil
        }
        return nil
}

func autoConvert_templates_Validation_To_v1beta1_Validation(in *templates.Validation, out *Validation, s conversion.Scope) error {
        if in.OpenAPIV3Schema != nil {
                in, out := &in.OpenAPIV3Schema, &out.OpenAPIV3Schema
                *out = new(v1.JSONSchemaProps)
                if err := v1.Convert_apiextensions_JSONSchemaProps_To_v1_JSONSchemaProps(*in, *out, s); err != nil {
                        return err
                }
        } else {
                out.OpenAPIV3Schema = nil
        }
        return nil
}
```

constraint/Makefile
```diff
@@ -28,24 +28,29 @@ test: generate
 install: manifests
 	kubectl apply -f config/crds
 
+# Install the generation dependencies on the location machine
+gen-dependencies:
+	GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0
+	GO111MODULE=on go get k8s.io/code-generator/cmd/conversion-gen@v0.21.0
```
sozercan, line 34: should we align this with controller-gen dependency version (v0.20.2)?

maxsmythe: We probably want to align this with the cert-controller dependency version (but that is also v0.20.2):

https://github.com/open-policy-agent/cert-controller/blob/c6f841d912f7d057a4e9fc3fe2c49951b68ba931/go.mod#L9-L11

julianKatz: Will update

constraint/pkg/apis/templates/v1beta1/zz_generated.conversion.go
julianKatz: There's already a k8s bug for this (kubernetes/kubernetes#98380)

maxsmythe:
Checks are failing.

Can we exclude `zz_generated.*` files from the linter? Because we don't control their contents we're mostly concerned with whether or not they build.

For the unit tests, it looks like we need to change the controller-gen invocation to `go run ...`

sozercan:
@maxsmythe linter is failing because of undefined function getting generated (which is meant to cause build failure) https://github.com/open-policy-agent/frameworks/pull/111#pullrequestreview-647303629

constraint/pkg/apis/templates/v1beta1/zz_generated.conversion.go
julianKatz: False alarm!  I just needed to make that package available to `conversion-gen`.  Thanks for posting that example @sozercan, that was the clue I needed.

sozercan: Issue says it's for `types containing an array of pointers to custom objects.` Does this apply to our types?

julianKatz:
With the default call to `conversion-gen`, there are four missing conversions, divided into two packages (`constraint/pkg/apis/templates/v1alpha1`, `constraint/pkg/apis/templates/v1beta`).  The missing conversions are the same two:

```
apiextensions.JSONSchemaProps to apiextensionsv1beta1.JSONSchemaProps
```

and

```
 apiextensions.JSONSchemaProps to v1beta1.JSONSchemaProps
```

The first is fixed by including the `apiextensionsv1beta1` package (containing the auto-genned conversion function) as a flag on `converison-gen`:

```
--extra-peer-dirs=k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1
```

But, the second remains unfixed.  Why?

It turns out that the conversion function we'd expect is actually a human written one, found at `vendor/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1/conversion.go` (as opposed to in the generated `zz_generated.conversion.go` within the same package).  For some reason, `conversion-gen` isn't able to handle that when including `apiextensions/v1beta`.

AFAICT, this is a bug.

In the meantime, I can change the line of code manually and we can move on.  I'll file a bug in `conversion-gen` as well.

sozercan:
@julianKatz there are more investigations at https://github.com/kubernetes/kubernetes/issues/101567

constraint/pkg/apis/templates/v1beta1/zz_generated.conversion.go
julianKatz: Sorry, it isn't completely resolved.  I've included notes in `Makefile`.

But, I've got us frankensteined together enough to move forward.

julianKatz:
Update:

I've gotten around the `conversion-gen` bug with some manual code editing and created an issue (#112) to track the fix in this repository.

In the meantime, I believe this PR can move forward.

julianKatz:
Unit tests are passing now.  PTAL: @maxsmythe @sozercan @shomron

constraint/Makefile
```diff
@@ -21,31 +22,42 @@ docker-internal-test:
...
 # Hook to run docker tests
-test: generate
+test:
 	docker build . -t constraint-test && docker run -t constraint-test
 
 # Install CRDs into a cluster
 install: manifests
 	kubectl apply -f config/crds
 
+# Install the generation dependencies on the location machine
```
maxsmythe, line 32: nit: location => local?

maxsmythe (approved):
Still LGTM with 1 nit, @sozercan LGTY?

sozercan:
Not a huge fan of the manual code update, but I don't see any other way forward :( Can we document what to manually update in conversion? (perhaps add in #112)

sozercan (approved):

constraint/Makefile
julianKatz: oops, will change

julianKatz:
@sozercan Good call.

I'll add a sample diff in #112 and then link to their from both of the conversion files

sozercan:
@julianKatz ready to merge?

PR #113, opened 2021-05-04
Add x-kubernetes-preserve-unknown-fields during CT conversion

julianKatz:
v1 CRDs require that any JSONSchema information within them be
"structural", as defined in

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

Gatekeeper Constraint Templates (current supported versions v1alpha1,
v1beta1) contain information that is inserted into a CRD. That
information must be structural to be inserted into a v1 CRD.

As CTs are written by users, their `validation` section will contain
JSONSchema information that we as developers cannot control.  Further,
users already have v1alpha1 and v1beta1 CTs running on production
clusters that are expected to work as they upgrade gatekeeper and
their k8s cluster.

As v1beta1 CRDs did _not_ have the requirement of structural schemas,
v1beta1 CTs also do not.  Thus, our users could be passing in
non-structural information that we will have to fit into a v1 CRD.

To bridge this gap, this PR includes custom conversion functions for
v1alpha1 and v1beta1 CT `valiation` sections into the internal,
versionless CT representation.  These conversion functions transform the
input information, adding the `x-kubernetes-preserve-unknown-fields:
true` key/value pair where necessary.  This prevents the API server from
pruning fields that do not have the necessary type information, as it
would do were we to avoid this change.  This satisfies the definition of
a structural schema as defined in the aforementioned docs page.

This PR also includes a few missing `type: object` declarations in the
schema chunks it provides to Gatekeeper.


constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,63 @@
...
+// AdAAddPreserveUnknownFields recurses through an *apiextensionsv1beta1.JSONSchemaProps
+// data structure, adding `x-kubernetes-preserve-unknown-fields: true` at every level
+// that type is equal to "object", "array", or is undefined.
+func AddPreserveUnknownFields(sch *apiextensionsv1beta1.JSONSchemaProps) error {
+	// Both an object and an array can have values not described in the schema.
+	// A blank Type could be anything, including an object or array, so we should
+	// treat it the same as those types.
+	if sch.Type == "" || sch.Type == "object" || sch.Type == "array" {
+		trueBool := true
+		sch.XPreserveUnknownFields = &trueBool
```
maxsmythe, line 31: If x-preserve-unknown-fields is explicitly set to `false`, we probably want to preserve that (b/c it's a pointer I'm hoping we can tell the difference between undefined and intentionally false)

Also, no need to set XPreserveUnknown for type "array", there we just need to make sure sch.Items is structural if it's an Array.

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,69 @@
...
+				return err
+			}
+
+			// As v is not a pointer, we need to set the contents of v back into the original data structure
+			sch.Properties[k] = v
+		}
+	}
+
+	if sch.Items != nil {
+		if sch.Items.Schema != nil {
```
maxsmythe, line 46: Do we know the rules for when Items.Schema vs Items.JSONSchemas are set?

We should require at least one of them to be set in order to satisfy structural schema. It looks like the spec requires `items` to be set, though maybe there's a subtlety I'm missing:

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

constraint/pkg/apis/templates/transform_test.go
```diff
@@ -0,0 +1,355 @@
...
+				Type:                   "object",
+			},
+		},
+		{
+			name: "type array",
+			v: &apiextensionsv1beta1.JSONSchemaProps{
+				Type: "array",
+			},
+			exp: &apiextensionsv1beta1.JSONSchemaProps{
+				XPreserveUnknownFields: &trueBool,
```
maxsmythe, line 66: This should have a schema defined for `items`, per the spec:

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

constraint/pkg/apis/templates/transform_test.go
```diff
@@ -0,0 +1,355 @@
...
+		{
+			name: "map with empty JSONSchemaProps value",
+			v: &apiextensionsv1beta1.JSONSchemaProps{
+				Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
+					"foo": {},
+				},
+			},
+			exp: &apiextensionsv1beta1.JSONSchemaProps{
+				XPreserveUnknownFields: &trueBool,
+				Properties: map[string]apiextensionsv1beta1.JSONSchemaProps{
```
maxsmythe, line 79: Shall we also test if preserve-unknown is specifically set to false?

constraint/pkg/apis/templates/transform_test.go
```diff
@@ -0,0 +1,355 @@
...
+					Schema: &apiextensionsv1beta1.JSONSchemaProps{
+						AdditionalProperties: &apiextensionsv1beta1.JSONSchemaPropsOrBool{
+							Allows: false,
+						},
+					},
+				},
+			},
+			exp: &apiextensionsv1beta1.JSONSchemaProps{
+				XPreserveUnknownFields: &trueBool,
+				AdditionalProperties: &apiextensionsv1beta1.JSONSchemaPropsOrBool{
```
maxsmythe, line 251: Per the spec, AdditionalProperties just shouldn't be set:

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,69 @@
...
+			if err := AddPreserveUnknownFields(&v); err != nil {
+				return err
+			}
+
+			// As v is not a pointer, we need to set the contents of v back into the original data structure
+			sch.Items.JSONSchemas[k] = v
+		}
+	}
+
+	if sch.AdditionalProperties != nil && sch.AdditionalProperties.Schema != nil {
```
maxsmythe, line 62: technically this should not be set, per the spec:

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

So if it exists, this is already broken. Fortunately, I think they never allowed AdditionalProperties to be set? (hopefully...)

julianKatz: AFAICT `additionalProperties` has the following restrictions on it (from that post):

> 3. does not set... additionalProperties... within an allOf, anyOf, oneOf or not, with the exception of the two pattern for x-kubernetes-int-or-string: true (see below).

And from lower in the article:

> The field additionalProperties cannot be set to false.
> The field additionalProperties is mutually exclusive with properties.

That makes it seem like `additionalProperties` _can_ be set, but not inside of those other validating directives like `allOf` and not alongside a `properties`.

Am I interpreting that correctly?

constraint/pkg/apis/templates/transform_test.go
julianKatz: Good catch.  Will do now

constraint/pkg/apis/templates/transform.go
maxsmythe: Ah, you are, I missed that.

constraint/pkg/apis/templates/transform_test.go
julianKatz: We resolved this in [another thread](https://github.com/open-policy-agent/frameworks/pull/113#discussion_r626039744)

constraint/pkg/apis/templates/transform_test.go
julianKatz: While this makes total sense, I'm not sure what in the spec actually spells out that `type: array` will fail if not accompanied by `items`.  There is this quote: 

>specifies a non-empty type (via type in OpenAPI) for the root, for each specified field of an object node (via properties or additionalProperties in OpenAPI) and for each item in an array node (via items in OpenAPI), with the exception of:
> - a node with x-kubernetes-int-or-string: true
> - a node with x-kubernetes-preserve-unknown-fields: true

But this seems to say that an _item_ in an array node needs to have a type or `x-kub...`, but not that `items` needs to exist.

I did some searching through the [OpenAPI Spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md) for `items` and `array` to look for a clue but did not find one.  Is there something that you've seen?

The best I can find is on json-schema.org, in the [Understanding JSON Schema page](https://json-schema.org/understanding-json-schema/reference/array.html#items):

> By default, the elements of the array may be anything at all. However, it’s often useful to validate the items of the array against some schema as well. This is done using the items, additionalItems, and contains keywords.

Overall, it feels a little ambiguous to me. But, I'm happy to follow if you feel you understand what's implied.

constraint/pkg/apis/templates/transform.go
julianKatz: That makes sense, will update.

constraint/pkg/apis/templates/transform.go
julianKatz: JSONSchema allows one to set multiple different schemas under `items` as a way of providing per-field [tuple validation](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation).  

I'm not seeing anything in the k8s spec that overtly disallows this, but I did some digging with SourceGraph and I found that the api server runs this [validate()](https://github.com/kubernetes/apiextensions-apiserver/blob/7537a765248141dadcfd216ca03c0a09260c3ae4/pkg/apis/apiextensions/validation/validation.go#L997) function at each as it [recurses through a CRD OpenAPI Schema](https://github.com/kubernetes/apiextensions-apiserver/blob/7537a765248141dadcfd216ca03c0a09260c3ae4/pkg/apis/apiextensions/validation/validation.go#L740).  This logic [disallows JSONSchemas](https://github.com/kubernetes/apiextensions-apiserver/blob/7537a765248141dadcfd216ca03c0a09260c3ae4/pkg/apis/apiextensions/validation/validation.go#L1049), requiring that its length be zero.

```golang
if schema.Items != nil && len(schema.Items.JSONSchemas) != 0 {
	allErrs = append(allErrs, field.Forbidden(fldPath.Child("items"), "items must be a schema object and not an array"))
}
```

constraint/pkg/apis/templates/transform_test.go
maxsmythe: From line item 1 for specifying a structural schema:

`and for each item in an array node (via items in OpenAPI),`

julianKatz: Some comments from chatting with max:

- if `type: array` and no `items`, add `items` with `x-kub...` inside of it
- if `type: array`, don't put `x-kub...` at that level
- if no `type: array`, add `x-kub...`, even if there is `items`

constraint/pkg/apis/templates/transform.go
julianKatz: Max and I discussed and have decided that, since JSONSchemas is not supported by the API server (and wasn't as far back as k8s 1.14), we shouldn't take on supporting it in our code.  Instead, we'll return an error if we find JSONSchemas to be non `nil`.

maxsmythe (approved):
LGTM

@sozercan @ritazh @shomron LGTY?

sozercan (commented):
Have you tried this in Gatekeeper? I am seeing 


`$ kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/allowedrepos/template.yaml`

`Error from server ([spec.validation.openAPIV3Schema.properties[metadata].type: Invalid value: "": must be object, spec.validation.openAPIV3Schema.properties[metadata].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[kinds].items.properties[apiGroups].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[kinds].items.properties[kinds].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[kinds].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[labelSelector].properties[matchExpressions].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[labelSelector].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[namespaceSelector].properties[matchExpressions].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[namespaceSelector].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root]): error when creating "https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/allowedrepos/template.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [spec.validation.openAPIV3Schema.properties[metadata].type: Invalid value: "": must be object, spec.validation.openAPIV3Schema.properties[metadata].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[kinds].items.properties[apiGroups].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[kinds].items.properties[kinds].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[kinds].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[labelSelector].properties[matchExpressions].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[labelSelector].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[namespaceSelector].properties[matchExpressions].items.type: Required value: must not be empty for specified array items, spec.validation.openAPIV3Schema.properties[spec].properties[match].properties[namespaceSelector].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].properties[match].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.properties[spec].type: Required value: must not be empty for specified object fields, spec.validation.openAPIV3Schema.type: Required value: must not be empty at the root]`

Sounds like we might need to add `type`?

julianKatz:
> Sounds like we might need to add `type`?

TL;DR: this is expected, b/c open-policy-agent/gatekeeper#1249 is unmerged

@sozercan I believe this failure is related to un-merged open-policy-agent/gatekeeper#1249.  That PR adds type information to the `match` section of a Constraint Kind.  

[1249](https://github.com/open-policy-agent/gatekeeper/pull/1249) is experiencing e2e failures where the webhook seems to interpret the schema section as a v1 CRD due to the presence of `x-kubernetes-preserve-unknown-fields` (even if the overall CRD is `v1beta1`).  As the creation of a structural Constraint kind CRD requires two structural sections (`parameters`, which we have updated in this PR, and `match`), it will error without both.  

To get a passing e2e suite, we'll probably need to update Gatekeeper to include the changes in this PR _as part of_ PR [1249](https://github.com/open-policy-agent/gatekeeper/pull/1249).  That way, both of the schema-based CRD ingredients will be structural at the same time.

Let me know if that seems off-base, or if I can provide additional clarification.

julianKatz:
TL;DR: This PR is fully ready.  I'm able to replace `frameworks` in gatekeeper with this version and all tests pass.

To test that this change was working correctly, I replaced the `frameworks/constraint` dependency in gatekeeper with my un-merged version.  I did so via a line like this in gatekeeper's `go.mod`:

```
replace github.com/open-policy-agent/frameworks/constraint => github.com/julianKatz/frameworks/constraint v0.0.0-20210513205434-f5b7f6a7c769
```

Until my most recent addition to the `addPreserveUnknownFields` function, I was seeing a certain class of error in the tests for gatekeeper's `readiness` package.  The ingestion of the `k8srequiredlabels` template was failing:

```
2021-05-13T22:08:27.653Z        ERROR   controller      CRD creation error      {
  "kind": "ConstraintTemplate",
  "process": "constraint_template_controller",
  "template_name": "k8srequiredlabels",
  "error": "spec.validation.openAPIV3Schema.properties[spec].properties[parameters].properties[labels].items: Required value: must be specified",
  "errorCauses": [
    {
      "error": "spec.validation.openAPIV3Schema.properties[spec].properties[parameters].properties[labels].items: Required value: must be specified"
    }
  ]
}
```

The error stated that `parameters.labels.items` needed to be specified.  But, it _already was_.  What could be happening?  I looked at the ConstraintTemplate used in the test.  The relevant section was:

```
      validation:
        # Schema for the `parameters` field
        openAPIV3Schema:
          properties:
            labels:
              type: array
              items: string
```

As we see here, the `items` field _is included_.  However, the `string` keyword is used incorrectly.  `items:` is supposed to be an object containing a schema for each array item.  If the type of that item should be a `string`, the section should like this:

```
labels:
  type: array
  items:
    type: string
```

As this schema (`string`) was invalid, it became `null` when ingested into gatekeeper.  Once converted to golang structs, `items: null` became:

```golang
&apiextensionsv1beta1.JSONSchemaProps{
    Type:  "array",
    Items: &apiextensionsv1beta1.JSONSchemaPropsOrArray{},
}
```

This empty `JSONSchemaPropsOrArray` was incorrectly interpreted by my algorithm to be a valid schema, and thus was left un-transformed.

The fix was to recognize this schema as invalid and overwrite it with:

```
Items: &apiextensionsv1beta1.JSONSchemaPropsOrArray{
	Schema: &apiextensionsv1beta1.JSONSchemaProps{
		XPreserveUnknownFields: &trueBool,
	},
},
```

With this implemented and `vendor`ed into gatekeeper, all gatekeeper unit tests pass.

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,84 @@
...
+	// typeWithFields := sch.Type == "" || sch.Type == "object"
+	typeWithFields := sch.Type == "" || sch.Type == "object"
+
+	// Don't override an already set XPreserveUnknownFields
+	if typeWithFields && sch.XPreserveUnknownFields == nil {
+		sch.XPreserveUnknownFields = &trueBool
+	}
+
+	if sch.Properties != nil {
+		for k, v := range sch.Properties {
```
briantkennedy, line 41: nit: to avoid the setting back onto original structure, use:

`AddPreserveUnknownFields(&sch.Properties[k])`

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,84 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package templates
+
+import (
+	"fmt"
+
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
sozercan, line 21: do we want this to be `v1`?

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,84 @@
...
+var trueBool = true
+
+// AddPreserveUnknownFields recurses through an *apiextensionsv1beta1.JSONSchemaProps
+// data structure, adding `x-kubernetes-preserve-unknown-fields: true` at every level
+// that type is equal to "object", "array", or is undefined.
+func AddPreserveUnknownFields(sch *apiextensionsv1beta1.JSONSchemaProps) error {
+	// An object can have values not described in the schema.  A blank Type could be
+	// anything, including an object, so we should treat it the same as those types.
+	// typeWithFields := sch.Type == "" || sch.Type == "object"
+	typeWithFields := sch.Type == "" || sch.Type == "object"
```
briantkennedy, line 33: I'd be more comfortable with top level logic that switched based on sch.Type then handled as appropriate.  This looks like several functions in one since each if block seems independent.

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,84 @@
...
+		sch.Items = &apiextensionsv1beta1.JSONSchemaPropsOrArray{
+			Schema: &apiextensionsv1beta1.JSONSchemaProps{
+				XPreserveUnknownFields: &trueBool,
+			},
+		}
+
+		return nil
+	}
+
+	if sch.Items != nil {
```
briantkennedy, line 65: does this logic need to be inside a `sch.Type == "array"` check?

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,84 @@
...
+				return err
+			}
+		}
+
+		if sch.Items.JSONSchemas != nil {
+			return fmt.Errorf("non-nil JSONSchemas encountered, multiple schemas are not supported")
+		}
+	}
+
+	if sch.AdditionalProperties != nil && sch.AdditionalProperties.Schema != nil {
```
briantkennedy, line 77: Is this valid for Arrays, or just Objects?

constraint/pkg/apis/templates/transform.go
julianKatz: Both `v1alpha1` and `v1beta1` ConstraintTemplate versions use `apiextensionv1beta1`, so that's where we need support.

`v1` ConstraintTemplate can use the new version, and there we won't be doing any of this kind of transformation.

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,84 @@
...
+// that type is equal to "object", "array", or is undefined.
+func AddPreserveUnknownFields(sch *apiextensionsv1beta1.JSONSchemaProps) error {
+	// An object can have values not described in the schema.  A blank Type could be
+	// anything, including an object, so we should treat it the same as those types.
+	// typeWithFields := sch.Type == "" || sch.Type == "object"
+	typeWithFields := sch.Type == "" || sch.Type == "object"
+
+	// Don't override an already set XPreserveUnknownFields
+	if typeWithFields && sch.XPreserveUnknownFields == nil {
+		sch.XPreserveUnknownFields = &trueBool
```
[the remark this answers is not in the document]
julianKatz: Good call.  Fixing.

constraint/pkg/apis/templates/transform.go
julianKatz: That doesn't seem to compile:

```
invalid operation: cannot take address of sch.Properties[k] (map index expression of type v1beta1.JSONSchemaProps)
```

[Stackoverflow says](https://stackoverflow.com/a/32496031) this is actually a design feature of golang that's made to prevent the creation of a dangling pointer.

constraint/pkg/apis/templates/transform.go
julianKatz: It's possible for `sch.Items` to be defined while no `Type` is defined.  Basically the user can define _anything_ in this schema area, so we can't depend on the relationships between fields that we'd usually expect.

constraint/pkg/apis/templates/transform.go
julianKatz: `AdditionalProperties` [is meant to provide some schema validation for fields not defined in `properties`](https://json-schema.org/understanding-json-schema/reference/object.html#index-1:~:text=The%20additionalProperties%20keyword%20is%20used%20to,if%20they%20are%20each%20a%20string%3A).  So it's really just another way of defining an object.

constraint/pkg/apis/templates/transform.go
julianKatz: I think I understand what you mean, but I don't think the blocks are quite as independent as they may seem.  The reason I say that is because we cannot rely on `Type` being defined.  It's very often left undefined by our users, or at least included inconsistently.

For example, all of these are possible in a given schema struct:

- `properties` with `type: object`
- `properties` without `type:object`
- `items` with `type: array`
- `items` without `type: array`
- `type: object` with no `properties`
- `type: array` with no `items`

I've refactored some of the function with a `switch` statement.  That section decides whether-or-not we add `x-kub...`, and is based on `Type`.  

The latter half of the function (diving into `properties`, `items`, `additionalProperties`) can't depend on the un-reliably defined `Type`, and so I've left out outside of the `switch`.

LMK your thoughts.

constraint/pkg/apis/templates/transform.go
julianKatz: I did some testing and it turns out you cannot define `items:` with any type other than `array`.  Also, if you have `type: array`, `items:` _must_ be defined or the CRD is not accepted by the API server.

That clarifies the relationship a little, but doesn't save us much in terms of code complexity AFAICT.

constraint/pkg/apis/templates/transform.go
```diff
@@ -0,0 +1,83 @@
...
+// AddPreserveUnknownFields recurses through an *apiextensionsv1beta1.JSONSchemaProps
+// data structure, adding `x-kubernetes-preserve-unknown-fields: true` at every level
+// that type is equal to "object", "array", or is undefined.
+func AddPreserveUnknownFields(sch *apiextensionsv1beta1.JSONSchemaProps) error {
+	switch sch.Type {
+	// An object can have values not described in the schema.  A blank Type could be anything,
+	// including an object, so we add x-kubernetes-preserve-unknown-fields: true to both.
+	case "":
+		fallthrough
+	case "object":
```
briantkennedy, line 33: nit: `case "", "object":` is more ideal.

constraint/pkg/apis/templates/transform.go
briantkennedy: I guess there's still some things in go that will surprise me.  Perhaps use a temp variable for the pointer instead?

```
for k := range ... {
  v := &sch.Properties
}
```

constraint/pkg/apis/templates/transform.go
briantkennedy: This is an interesting question.  If the user previously specified `.Items` in an invalid context, was the API server throwing an error?

constraint/pkg/apis/templates/transform.go
briantkennedy: Was it previously invalid for the user to specify `.AdditionalProperties` for a non-object type?  If so, this could be moved to the object handling portion.

briantkennedy (commented):
The only thing I'm still wondering about is how the API server used to handle malformed api spec.  Was it very lenient and let people specify nonsense in v1beta1, or did it reject things that were incorrect?  I think this makes a difference because there could be a lot of broken OpenAPI specs out there, and the right thing might actually be to remove the provided schema require the user to flip an `enableCRDv1SchemaChecking` flag on each template for it to be enforced.

sozercan (approved):
LGTM, pending comments from @briantkennedy

constraint/pkg/apis/templates/transform.go
julianKatz: The thing that makes me not like splitting the latter half of this function into a type-based switch statement is that all of the logic needs to be called for `Type: ""` and only parts of it are called for `Type: "object"` and `Type: "array"`.  If I'm to avoid repeating code, then I split the various parts of this logic into separate functions.  These functions would then call back recursively to `AddPreserveUnknownFields()`.  Personally I find that more confusing than what I have now.

Is there something I'm not seeing about how this could be put together?

constraint/pkg/apis/templates/transform.go
julianKatz: What exactly do you mean by an "invalid context"?

`items: ` can be declared alongside `type: array` or with no `type` in a v1beta1 CRD schema.  Any other type (`string`, `object`, etc.) yields an error from the API server.

briantkennedy: By invalid context, I mean `type`.

My question here is in terms of correct behavior, if some combination of `items` and `properties`/`additionalProperties` are present, what should this do?

julianKatz: Upon doing more testing, I'm seeing that the following is (unfortunately) _allowed_ in a `v1beta1 ConstraintTemplate`:

```
      validation:
        # Schema for the `parameters` field
        openAPIV3Schema:
          properties:
            foo:
              type: string
            bar:
              type: string
          items:
            type: string
```

My takeaway from this is that there is no XOR between these `properties`, `items`, `additionalProperties` expected in `v1beta1 ConstraintTemplate`.

constraint/pkg/apis/templates/transform.go
julianKatz: Good call. done.

constraint/pkg/apis/templates/transform_test.go
```diff
@@ -0,0 +1,330 @@
...
+			},
+			exp: &apiextensionsv1beta1.JSONSchemaProps{
+				XPreserveUnknownFields: &trueBool,
+				Properties:             nil,
+			},
+		},
+		{
+			name: "x-kubernetes-preserve-unknown-fields is present already but false",
+			v: &apiextensionsv1beta1.JSONSchemaProps{
+				XPreserveUnknownFields: &falseBool,
```
briantkennedy, line 55: `false` appears to not be invalid for this field.

https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types_jsonschema.go#L95

https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1beta1/types_jsonschema.go#L95

constraint/pkg/apis/templates/transform.go
julianKatz: This yields the same error:

```
		for k := range sch.Properties {
			v := &sch.Properties[k] // invalid operation: cannot take address of sch.Properties[k]...
			if err := AddPreserveUnknownFields(v); err != nil {
				return err
			}
		}
```

constraint/pkg/apis/templates/transform_test.go
julianKatz: Ah, nice.  We can take this out @maxsmythe

constraint/pkg/apis/templates/v1alpha1/conversion.go
```diff
@@ -0,0 +1,42 @@
...
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/conversion"
+)
+
+func Convert_v1alpha1_Validation_To_templates_Validation(in *Validation, out *coreTemplates.Validation, s conversion.Scope) error { //nolint:golint
+	if in.OpenAPIV3Schema != nil {
+		in, out := &in.OpenAPIV3Schema, &out.OpenAPIV3Schema
+
+		if err := apisTemplates.AddPreserveUnknownFields(*in); err != nil {
```
maxsmythe, line 30: It looks like conversion functions are written in such a way as to avoid modifying the `in` value. I think the quickest way to have the same effect here is to have:

`in := in.OpenAPIV3Schema.DeepCopy()`

constraint/pkg/apis/templates/v1beta1/conversion.go
```diff
@@ -0,0 +1,42 @@
...
+	apisTemplates "github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates"
+	coreTemplates "github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	"k8s.io/apimachinery/pkg/conversion"
+)
+
+func Convert_v1beta1_Validation_To_templates_Validation(in *Validation, out *coreTemplates.Validation, s conversion.Scope) error { //nolint:golint
+	if in.OpenAPIV3Schema != nil {
+		in, out := &in.OpenAPIV3Schema, &out.OpenAPIV3Schema
```
maxsmythe, line 28: Same deepcopy comment

constraint/pkg/apis/templates/v1alpha1/conversion.go
julianKatz: Good call.  Will change.

PR #117, opened 2021-05-18
Preserve unknown fields in ByPodStatus

julianKatz:
A Gatekeeper e2e test is currently failing in
open-policy-agent/gatekeeper#1286 due to the use of the `operations`
field.  This field is added by Gatekeeper, under a ConstraintTemplate's
`byPodsStatus`.  This field relied on the v1beta1 CRD behavior, where
unknown fields are preserved by default.

As this field fundamentally has to do with gatekeeper, it's better not
to define it in the Constraint Framework.  Other CF consumers have no
need for the Operations field.

Even better, we may eventually define ConstraintTemplateSpec in
Constraint Framework consumers, rather than in the library itself.

For now, adding `x-kubernetes-preserve-unknown-fields: true` to
ByPodStatus will solve our problem and maintain the existing behavior.

### DIFF
--- a/constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
+++ b/constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
@@ -197,6 +197,7 @@ spec:
                       description: a unique identifier for the pod that wrote the
                         status
                       type: string
+                      x-kubernetes-preserve-unknown-fields: true
                     observedGeneration:
                       format: int64
                       type: integer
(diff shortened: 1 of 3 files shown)


maxsmythe (approved):
LGTM

maxsmythe:
@ritazh @shomron @sozercan this PR is blocking for handling the v1beta1 CRD deprecation, it fixes an e2e failure in this PR:


https://github.com/open-policy-agent/gatekeeper/pull/1286

sozercan (approved):

PR #119, opened 2021-05-20
Upgrade the CF client to apiextensionsv1

julianKatz:
This PR upgrades the Constraint Framework client to use the
apiextensionsv1 library for handling CRDs.  This is necessary for
supporting k8s 1.22, where apiextensionsv1beta1 CRDs are removed.

Crucially, this change switches CF to use the v1 CRD defaulting
functions, which do not set Spec.PreserveUnknownFields.  This field set
to true by default in v1beta1 CRD, but is disallowed in v1 CRD.  This
was preventing me from doing the same apiextensionsv1beta1 -->
apiextensionsv1 upgrade in Gatekeeper.


maxsmythe (approved):
LGTM

ritazh (approved):

PR #120, opened 2021-05-21
Add x-preserve... to Constraint kind status

julianKatz:
Gatekeeper adds the byPod field to the status of a Constraint once it is
applied to the cluster.  This previously relied on the preservation of
unknown fields by the API server.  As we are now applying Constraint kind
CRDs as v1 CRDs, this behavior breaks.

This PR adds an empty `status` to the Constraint kind's CRD's schema.
That status includes the `x-kubernetes-preserve-unknown-fields: true`
key/value pair, retaining the existing Gatekeeper functionality.


maxsmythe (approved):
LGTM

ritazh (approved):
Thanks for spotting this!

PR #121, opened 2021-06-01
Add ConstraintTemplate v1

julianKatz:
We recently upgraded Constraint Framework to produce v1 CRDs when
creating Constraint kind CRDs.  This was in preparation for the release
of k8s 1.22, which removes the `v1beta1` CRD version.  See
open-policy-agent/gatekeeper#550 for more info.

As v1beta1 ConstraintTemplate did _not_ required any user-entered schema
information to be structural, transformation logic was implemented to
"structuralize" the user-inputted schema information as needed.

The new v1 ConstraintTemplate version purposefully does _no_
transformation, as it is meant to put the ConstraintTemplate creation
experience in line with that of a v1 CRDs.  Any schema information added
by the user is expected to be structural.  If non-structural schema info
is added, an error should be returned.


constraint/pkg/apis/templates/v1/constrainttemplate_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+type Names struct {
+	Kind       string   `json:"kind,omitempty"`
+	ShortNames []string `json:"shortNames,omitempty"`
+}
+
+type Validation struct {
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:validation:Type=object
+	// +kubebuilder:pruning:PreserveUnknownFields
```
maxsmythe, line 49: We should add another bool field to this object (for all versions) called `legacySchema`.

For `v1beta1` and `v1alpha1` it should default to `true`, for `v1` it should default to `false`.

If `legacySchema` is `true`, then we perform x-preserve injections when creating the versionless object.

Doing things this way means that all versions remain wire-compatible and we have no need to write a conversion webhook.

julianKatz: done @maxsmythe

julianKatz:
@maxsmythe and I have been trying to decide if Gatekeeper needs a [Version Conversion Webhook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion) to handle the conversion between `v1alpha1`, `v1beta1`, and `v1` ConstraintTemplates.

To deal with reverse-compatibility, we've added the `legacySchema: bool` field.  This handles the case of requesting a CT that was applied as `v1beta1` as a `v1`, where the schema is expected to be structural.  Rather than transform the schema information at request time (which would also require that we _un_-transform it when going back the other way), we can slap on`legacyScheam: true` and leave it as is.  The difference between the versions will be that `legacySchema` defaults to `true` in `v1alpha1` and `v1beta`, but to `false` in `v1`.

To validate that this acted as expected, I did the following manual test:

1. build and deploy Gatekeeper at HEAD, with a version of the CT CRD that contains only versions `v1alpha1` and `v1beta1`
1. Apply a v1beta1 CT, wait for it to persist on the API server
1. Apply the _updated_ CT CRD, which contains version `v1`.  `v1` becomes the storage version.
1. Use kubectl to retrieve the CT we applied earlier

This test showed that the CT (originally applied as a `v1beta1`) was retrieved by kubectl as `v1`, and that `legacySchema` was set to `true`.  This bodes well.  After an upgrade, it looks like Kubernetes will correctly assign our defaults to the object in a way that `constrainttemplate_controller.go` can depend on.

Now, I'm going to dig into the details of how this process happens in the API server.  We want to insure that this is the supported path, and not just a [happy accident](https://media1.giphy.com/media/Q8vZZ6kGS1VFm/giphy.gif?cid=ecf05e472i5e4r15ibnodmjn7gbtqwtpmqyc94uuadt7ww3y&rid=giphy.gif&ct=g).

julianKatz:
Further investigation in the docs has confirmed this behavior:

"Defaulting happens on the object... [when reading from etcd using the storage version defaults](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting:~:text=when%20reading%20from%20etcd%20using%20the%20storage%20version%20defaults)".  

As the [previous comment](https://github.com/open-policy-agent/frameworks/pull/121#issuecomment-854033558) showed, the defaults from the version that a given CR was stored as are applied when retrieving that CR.  This appears to be the case even when that CR is retrieved another version.  In my test case, that meant that a `v1beta1` ConstraintTemplate saved in etcd was given (at retrieval time) the default value assigned to v1beta1 CTs in the CT CRD.  This happened even though the CT was requested at the newer `v1` version, which has a _different_ default value.

maxsmythe:
What's here LGTM, but should we also add the honoring of legacySchema to this PR so that we don't recognize v1 without also recognizing its behavioral changes?

maxsmythe (changes requested):
Forgot to mark my comment as a change request on the PR

julianKatz:
> What's here LGTM, but should we also add the honoring of legacySchema to this PR so that we don't recognize v1 without also recognizing its behavioral changes?

Does the newer version of the code do what you had in mind?

Basically, for all three CT versions, the transformation is only done if `legacySchema: true` is specified.

constraint/pkg/apis/templates/v1/constrainttemplate_types_test.go
```diff
@@ -0,0 +1,256 @@
...
+										"key":          {Type: "string"},
+										"allowedRegex": {Type: "string"},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			error: false,
```
maxsmythe, line 240: We should remove the "error" config if we're not going to use it.

maxsmythe (approved):
LGTM with one nit.

@ritazh @shomron @sozercan LGTY?

constraint/pkg/apis/templates/v1/constrainttemplate_types_test.go
julianKatz: I've removed this one and its equivalents from the transformation tests across the three CT version packages

julianKatz:
ping :) @shomron @sozercan

shomron (approved):
LGTM

PR #123, opened 2021-06-09
ConstraintTemplate versions use apiextensionsv1

julianKatz:
Previously, I left ConstraintTemplate (CT) versions v1alpha1 and v1beta1
using apiextensionsv1beta1.JSONSchemaProps.  I thought this appropriate,
given that was what they were previously implemented as.  v1 CT, on the
other hand, would use v1beta1 JSONSchemaProps.

This turns out to be a bad idea.

Any logic we write that interacts with OpenAPIV3Schema in a CT's
Validation section is forced to interact with OpenAPIV3Schema's type.
Before this PR, that type is apiextensionsv1beta1.JSONSchemaProps.  This
makes our life difficult, as v1 CT's version of this variable will have
a different type.  This would then require us to re-implement any logic
for this section to implement a second type, or to do some gross
conversions.

Further, apiextensionsv1beta1 will be removed from
k8s.io/apiextensions-apiserver, preventing us from upgrading that
library without doing this work.

Basically, this change is unavoidable and appropriate.

This blocks open-policy-agent/frameworks#121

### DIFF
--- a/constraint/Makefile
+++ b/constraint/Makefile
@@ -60,4 +60,4 @@ generate:
 		--input-dirs "./pkg/apis/templates/..." \
 		--go-header-file=./hack/boilerplate.go.txt \
 		--output-file-base=zz_generated.conversion \
-		--extra-dirs=k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1
+		--extra-dirs=k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1
--- a/constraint/pkg/apis/templates/transform.go
+++ b/constraint/pkg/apis/templates/transform.go
@@ -18,13 +18,13 @@ package templates
 import (
 	"fmt"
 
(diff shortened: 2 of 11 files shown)


maxsmythe (approved):
you ran `make generate` and there were no manifest changes?

If so, LGTM

@shomron @ritazh @sozercan LGTY?

ritazh (approved):

ritazh:
Getting an error running `make manifests` it could be my test env.

julianKatz:
> `make manifests` generated diff in the following files but they are all line break updates.
> modified: config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
> modified: deploy/crds.yaml

@ritazh What version of `controller-gen` are you running?  I just tried again and didn't see any diff.  Mine is:

```
❯ controller-gen --version
Version: v0.5.0
```

ritazh:
@julianKatz it's showing `Version: v0.5.0`

PR #124, opened 2021-06-10
Set PreserveUnknownFields: false in v1 CRDs

julianKatz:
Our CRDs were previously of version v1beta1.  In v1beta1,
spec.PreserveUnknownFields defaulted to `true`.  As we did not set this
value explicitly, the value was set to `true`.

We've now updated our CRDs to v1.  In v1, spec.PreserveUnknownFields
defaults to false.  In fact, it is being deprecated in favor of the more
targeted x-kubernetes-preserve-unknown-fields field.

Despite this new default, spec.PreserveUnknownFields remains `true` for
our CRDs, even after applying the new v1 versions.  This does not cause
any bad behavior, but is incorrect.  The simple fix is to set
spec.PreserveUnknownFields to false in all our CRDs.

As the Constraint kind CRD is generated in code, I've added this field
in constraint/pkg/client/crd_helpers.go.

The ConstraintTemplate CRD is generated using `controller-gen`, but 
v0.5.0 of the tool doesn't include an annotation for that, presumably
because it's not longer the supported path for disabling pruning.  So, I've
achieved the desired effect with a simple kustomize patch.

### DIFF
--- a/constraint/Makefile
+++ b/constraint/Makefile
@@ -40,7 +40,7 @@ manifests:
 		crd:allowDangerousTypes="true" \
 		paths="./pkg/..." \
 		output:crd:artifacts:config=config/crds
-	cp config/crds/* deploy/crds.yaml
+	kustomize build config/crds --output=deploy/crds.yaml
 
 lint:
 	golangci-lint -v run ./... --timeout 5m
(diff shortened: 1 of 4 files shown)


maxsmythe (commented):
Should we set this for the other CRDs as well?

maxsmythe (approved):
LGTM

@ritazh @shomron @sozercan LGTY?

maxsmythe:
@ritazh @shomron @sozercan poke

ritazh (approved):

PR #125, opened 2021-07-14
Use git lfs for large dependencies

willbeason:
This will prevent unnecessarily copying the large files unless needed, and get rid of the warning any time we run "git push" or similar.


willbeason:
This seems to no longer be required??

sozercan:
@willbeason due to https://github.com/open-policy-agent/opa/pull/3544?

willbeason:
Yeah, I hadn't realized when writing this that the upgrade PR was still open.

PR #129, opened 2021-07-29
Add crds.yaml-based defaulting functions to pkg/apis/templates/...

### LINKED ISSUE #128: Build a defaulting system that derives code from the ConstraintTemplate CRD yaml
ConstraintTemplate v1 introduced a new field: `legacySchema`.  This field is crucial for correctly interpreting (i.e. transforming) non-structural schemas in ConstraintTemplate resources, namely those that were declared as `v1beta1` CTs before structural schemas were introduced to kubernetes.

These defaults are already declared in the ConstraintTemplate CRD.  They are placed there during CRD yaml generation, and are derived from kubebuilder annotations that live in `constraint/pkg/apis/templates/v1/constrainttemplate_types.go` (for example, [that of legacySchema in CT v1](https://github.com/open-policy-agent/frameworks/blob/master/constraint/pkg/apis/templates/v1/constrainttemplate_types.go#L51)).

For our golang logic to have these same defaults, we are required to write custom defaulting functions and include them in each api version package.  This pattern requires that a new defaulting function be written each time a default is added.  This toil will undoubtedly lead to mistakes and future bugs.

To remedy this, we should make a pipeline that saves the generated CT CRD schema as a string constant in the golang code.  Generalized defaulting functions can be written that ingest this information and default values accordingly, just as [is done in the API server](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apiserver/schema/defaulting/algorithm.go#L34).

julianKatz:
Each ConstrainTemplate version (v1alpha1, v1beta1, v1) has a default
value for the LegacySchema field.  `true` for v1alpha1, v1beta1, and
`false` for v1.  These defaults are defined in the CRD yaml.

For the ConstraintTemplate golang types to have the same functionality
to have the same defaulting functionality.  We are required to write
custom defaulting functions for each defaulted field.

While we could do this, we could easily see how this practice could
become unsustainable.  A forgotten defaulting function would yield
unexpected bugs.

To remedy this, this PR adds defaulting functions that default based on
the content of deploy/crds.yaml, i.e. the ConstraintTemplate CRD
generated by controller-gen.

This provides us with a single defaulting source of truth: the
constrainttemplate_types.go files, which contain annotations that
define a field's default values.

Fixes #128


constraint/pkg/apis/templates/crd_schema.go
```diff
@@ -0,0 +1,60 @@
...
+	ConstraintTemplateSchemas = make(map[string]*schema.Structural)
+
+	// Ingest the constraint template CRD for use in defaulting functions
+	crdJSON, err := yaml.YAMLToJSONStrict([]byte(constraintTemplateCRDYaml))
+	if err != nil {
+		panic(errors.Wrap(err, "Failed to convert Constraint Template yaml to JSON"))
+	}
+
+	unCRD := unstructured.Unstructured{}
+	unCRD.UnmarshalJSON(crdJSON)
```
maxsmythe, line 27: We can't unmarshal directly to `constraintTemplateCRD` and save the middle step of converting to `unstructured`?

constraint/pkg/apis/templates/crd_schema.go
```diff
@@ -0,0 +1,60 @@
...
+	if err := apiextensionsv1.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+	if err := apiextensions.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+
+	// Fill version map with Structural types derived from ConstraintTemplate versions
+	for _, crdVersion := range constraintTemplateCRD.Spec.Versions {
+		versionedSchemaCopy := crdVersion.Schema.OpenAPIV3Schema.DeepCopy()
```
maxsmythe, line 46: Do we need to deepcopy this? It seems like we're throwing it away after-the-fact anyway?

constraint/pkg/apis/templates/v1/defaults.go
```diff
@@ -0,0 +1,30 @@
...
+
+func addDefaultingFuncs(scheme *runtime.Scheme) error {
+	return RegisterDefaults(scheme)
+}
+
+func SetDefaults_ConstraintTemplate(obj *ConstraintTemplate) {
+	// turn the CT into an unstructured
+	un, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		// JULIAN - what do I do here?  These functions don't return errors
```
maxsmythe, line 19: I think erroring is the best we can do here, unfortunately.

We should make sure this code path gets hit in a unit test to avoid errors, but these should never error, as any golang struct backing a resource should be convertible to unstructured.

maxsmythe (approved):
LGTM with a couple nits

maxsmythe (changes requested):
Actually, digging a bit deeper, it looks like we have these defaulting functions (which is good), but they're not being applied anywhere.

This means issues like https://github.com/open-policy-agent/gatekeeper/issues/1458 will still occur.

constraint/pkg/apis/templates/crd_schema.go
julianKatz: It doesn't look like `CustomResourceDefinition` has an Unmarshal function.

[`apiextensions-apiserver/pkg/apis/apiextensions/v1/marshal.go`](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/marshal.go) _does_ have unmarshalling logic, but it only supports doing so into a `JSONSchemaPropsOrArray`.

One idea:

I did the unmarshalling + `FromUnstructured()` because it made it easier to dive into the `versions:` and populate my `ConstraintTemplateSchemas` map.  I could probably simplify this logic by diving straight to `versions` in the unstructured and unmarshalling those JSON blobs straight into `apiextensionsv1.JSONSchemaProps`.  I'd then convert to `apiextensions.JSONSchemaProps` (apiextensions has no unmarshalling methods, only the versioned sub-packages) and create the `Structural` type for the map.

constraint/pkg/apis/templates/crd_schema.go
julianKatz: Removed.

constraint/pkg/apis/templates/crd_schema.go
julianKatz: I tried this^ and it's more trouble than it's worth.  B/c I use multiple properties of a `version` in my loop, it pays to convert to a `CustomResourceDefinition{}`.

constraint/pkg/apis/templates/v1/defaults.go
julianKatz: Removed comments.  Paths are already hit in tests.

constraint/pkg/apis/templates/v1/constrainttemplate_types.go
```diff
@@ -49,7 +49,7 @@ type Validation struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	OpenAPIV3Schema *apiextensionsv1.JSONSchemaProps `json:"openAPIV3Schema,omitempty"`
 	// +kubebuilder:default=false
-	LegacySchema bool `json:"legacySchema,omitempty"`
+	LegacySchema *bool `json:"legacySchema,omitempty"` // *bool allows for "unset" state
```
maxsmythe, line 52: nit: add "which we need to apply appropriate defaults"

maxsmythe (changes requested):
Looking good!

I think the last piece would be a "conversion" helper function that knows to apply defaults before converting. This could be leveraged by `gator test`

julianKatz:
> Looking good!
> 
> I think the last piece would be a "conversion" helper function that knows to apply defaults before converting. This could be leveraged by `gator test`

Is there an existing place where we'll use this function?  That would help me know I'm making the right thing.

constraint/pkg/apis/templates/v1/constrainttemplate_types.go
julianKatz: done

maxsmythe (approved):
LGTM, thanks for figuring this out!

maxsmythe:
Though it looks like there are linting errors, and we need to re-bake the CRD yaml into the golang constant

constraint/pkg/apis/templates/v1/helpers.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+
+	"github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// ToVersionless runs defaulting functions and then converts the ConstraintTemplate to the
+// versionless api representation
+func (versioned *ConstraintTemplate) ToVersionless(scheme *runtime.Scheme) (*templates.ConstraintTemplate, error) {
```
maxsmythe, line 12: We shouldn't require users to pass a scheme, just make your own local scheme for the frameworks package, I think we already have one:

https://github.com/open-policy-agent/frameworks/blob/c000ec35322ee04a6bc0a04041416432dfea943c/constraint/pkg/client/crd_helpers.go#L83-L89

maxsmythe (changes requested):
Sorry, missed some things

constraint/pkg/apis/templates/v1/helpers.go
julianKatz: done

constraint/pkg/apis/templates/v1/helpers.go
```diff
@@ -1,21 +1,16 @@
...
 // versionless api representation
-func (versioned *ConstraintTemplate) ToVersionless(scheme *runtime.Scheme) (*templates.ConstraintTemplate, error) {
-	if scheme == nil {
-		return nil, fmt.Errorf("Cannot convert using nil scheme")
-	}
-
-	if !scheme.IsVersionRegistered(SchemeGroupVersion) {
-		return nil, fmt.Errorf("GroupVersion '%v/%v' not registered in scheme", SchemeGroupVersion.Group, SchemeGroupVersion.Version)
+func (versioned *ConstraintTemplate) ToVersionless() (*templates.ConstraintTemplate, error) {
+	scheme := runtime.NewScheme()
```
maxsmythe, line 11: Can we make scheme a global variable?

constraint/pkg/apis/templates/v1beta1/helpers.go
```diff
@@ -1,21 +1,16 @@
...
 // versionless api representation
-func (versioned *ConstraintTemplate) ToVersionless(scheme *runtime.Scheme) (*templates.ConstraintTemplate, error) {
-	if scheme == nil {
-		return nil, fmt.Errorf("Cannot convert using nil scheme")
-	}
-
-	if !scheme.IsVersionRegistered(SchemeGroupVersion) {
-		return nil, fmt.Errorf("GroupVersion '%v/%v' not registered in scheme", SchemeGroupVersion.Group, SchemeGroupVersion.Version)
+func (versioned *ConstraintTemplate) ToVersionless() (*templates.ConstraintTemplate, error) {
+	scheme := runtime.NewScheme()
```
maxsmythe, line 11: make scheme a global variable?

constraint/pkg/apis/templates/v1alpha1/helpers.go
```diff
@@ -1,21 +1,16 @@
...
 // versionless api representation
-func (versioned *ConstraintTemplate) ToVersionless(scheme *runtime.Scheme) (*templates.ConstraintTemplate, error) {
-	if scheme == nil {
-		return nil, fmt.Errorf("Cannot convert using nil scheme")
-	}
-
-	if !scheme.IsVersionRegistered(SchemeGroupVersion) {
-		return nil, fmt.Errorf("GroupVersion '%v/%v' not registered in scheme", SchemeGroupVersion.Group, SchemeGroupVersion.Version)
+func (versioned *ConstraintTemplate) ToVersionless() (*templates.ConstraintTemplate, error) {
+	scheme := runtime.NewScheme()
```
maxsmythe, line 11: Global variable

maxsmythe (changes requested):

maxsmythe (approved):
LGTM

julianKatz:
ping :) @shomron @ritazh @sozercan

constraint/pkg/apis/templates/v1beta1/defaults.go
```diff
@@ -0,0 +1,28 @@
...
+
+func addDefaultingFuncs(scheme *runtime.Scheme) error {
+	return RegisterDefaults(scheme)
+}
+
+func SetDefaults_ConstraintTemplate(obj *ConstraintTemplate) { //nolint:golint
+	// turn the CT into an unstructured
+	un, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		panic("Failed to convert v1 ConstraintTemplate to Unstructured")
```
willbeason, line 19: Is there a way to let these return error, or does that mess up the code generation logic?

constraint/pkg/apis/templates/v1/conversion.go
```diff
@@ -47,5 +48,7 @@ func Convert_v1_Validation_To_templates_Validation(in *Validation, out *coreTemp
 	} else {
 		out.OpenAPIV3Schema = nil
 	}
+
+	out.LegacySchema = in.LegacySchema
```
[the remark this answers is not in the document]
maxsmythe: good catch

constraint/pkg/apis/templates/v1beta1/defaults.go
maxsmythe: Unfortunately the expected function signature does not allow for returning an error

constraint/pkg/apis/templates/crd_schema.go
```diff
@@ -0,0 +1,46 @@
...
+	if err := json.Unmarshal(crdJSON, constraintTemplateCRD); err != nil {
+		panic(errors.Wrap(err, "Failed to unmarshal JSON into CT CRD"))
+	}
+
+	// Fill version map with Structural types derived from ConstraintTemplate versions
+	for _, crdVersion := range constraintTemplateCRD.Spec.Versions {
+		versionlessSchema := &apiextensions.JSONSchemaProps{}
+		err := Scheme.Convert(crdVersion.Schema.OpenAPIV3Schema, versionlessSchema, nil)
+		if err != nil {
+			panic(errors.Wrap(err, "Failed to convert JSONSchemaProps"))
```
ritazh, line 36: nit: maybe also print the crdVersion name?

constraint/pkg/apis/templates/crd_schema.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var ConstraintTemplateSchemas map[string]*schema.Structural
+
+func initializeCTSchemaMap() {
+	// Setup the CT Schema map for use in generalized defaulting functions
+	ConstraintTemplateSchemas = make(map[string]*schema.Structural)
+
+	// Ingest the constraint template CRD for use in defaulting functions
+	crdJSON, err := yaml.YAMLToJSONStrict([]byte(constraintTemplateCRDYaml))
```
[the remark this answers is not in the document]
julianKatz: nice!  This removes multiple lines

constraint/pkg/apis/templates/crd_schema.go
```diff
@@ -0,0 +1,46 @@
...
+	for _, crdVersion := range constraintTemplateCRD.Spec.Versions {
+		versionlessSchema := &apiextensions.JSONSchemaProps{}
+		err := Scheme.Convert(crdVersion.Schema.OpenAPIV3Schema, versionlessSchema, nil)
+		if err != nil {
+			panic(errors.Wrap(err, "Failed to convert JSONSchemaProps"))
+		}
+
+		structural, err := schema.NewStructural(versionlessSchema)
+		if err != nil {
+			panic(errors.Wrap(err, fmt.Sprintf("Failed to create Structural for ConstraintTemplate version %v", crdVersion.Name)))
```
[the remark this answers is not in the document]
julianKatz: Did this for all of the `panic()` calls

constraint/pkg/apis/templates/init.go
```diff
@@ -0,0 +1,6 @@
+package templates
+
+func init() {
```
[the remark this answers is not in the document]
julianKatz: The problem wasn't the inability to have multiple `init()` calls, but rather that `initializeCTSchemaMap()` uses the `Scheme` that's created in `initializeScheme()`.  As init functions are run [alphabetically](https://medium.com/@markbates/go-init-order-dafa89fcef22#2ad9:~:text=As%20you%20can%20see%20the%20init%20functions%20are%20loaded%20in%20order%20by%20the%20filename.), this caused a null-pointer exception.

Rather than fix this through naming, I figured the best way to was to call them in an intentional order like I did in this `init.go` file.  I'll add a comment to `init.go` explaining that choice.

Or, perhaps there's a pattern that would better suit this use-case? @willbeason

constraint/pkg/apis/templates/v1beta1/defaults.go
willbeason: Ah, that's disappointing.

constraint/pkg/apis/templates/test_fakes.go
```diff
@@ -0,0 +1,71 @@
+package templates
+
+import (
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+var (
+	VersionedIncompleteSchema = &apiextensionsv1.JSONSchemaProps{
```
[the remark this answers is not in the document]
julianKatz: Good call.  Changed all these variables to functions.

constraint/pkg/apis/templates/v1/conversion.go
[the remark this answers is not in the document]
julianKatz: Awesome.  I've added a comment detailing why we do this and duplicated the change across the various `conversion.go` files.

constraint/pkg/apis/templates/v1/defaults_test.go
```diff
@@ -0,0 +1,76 @@
...
+					Rego:   `package hello ; violation[{"msg": "msg"}] { true }`,
+				},
+			},
+		},
+	}
+
+	scheme.Default(ct)
+
+	if ct.Spec.CRD.Spec.Validation.LegacySchema == nil {
+		t.Fatalf("legacySchema: Got nil, wanted false")
```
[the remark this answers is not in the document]
julianKatz: Changed across three functions

constraint/pkg/apis/templates/crd_schema.go
julianKatz: done

constraint/pkg/apis/templates/init.go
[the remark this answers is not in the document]
willbeason: Is there a reason the contents of those two methods can't be in the same function? `initializeScheme` is rather short.

willbeason (approved):

ritazh (approved):

PR #130, opened 2021-07-30
Inject empty `parameters` field for legacy schemas

maxsmythe:
To handle legacy (pre-v1) CRDs, we inject
x-kubernetes-preserve-unknown-fields
in order to preserve the legacy behavior
that unknown fields were not pruned by the API
server.

Unfortunately, we are not handling the case
where no parameters are provided, which currently
means that parameters will be stripped from users' CRs.

This fixes that.

Fixes: https://github.com/open-policy-agent/gatekeeper/issues/1468

### DIFF
--- a/constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
+++ b/constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
@@ -19,18 +19,13 @@ spec:
   - name: v1
     schema:
       openAPIV3Schema:
-        description: ConstraintTemplate is the Schema for the constrainttemplates
-          API
+        description: ConstraintTemplate is the Schema for the constrainttemplates API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this
-              object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
             type: string
           metadata:
             type: object
@@ -51,6 +46,8 @@ spec:
                             type: array
                         type: object
                       validation:
+                        default:
+                          legacySchema: false
                         properties:
                           legacySchema:
                             default: false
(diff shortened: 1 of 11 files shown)


constraint/deploy/crds.yaml
```diff
@@ -137,6 +139,8 @@ spec:
                             type: array
                         type: object
                       validation:
+                        default:
+                          legacySchema: true
                         properties:
                           legacySchema:
                             default: true
```
julianKatz, line 146: How do these interact?  Do we need both?

constraint/pkg/apis/templates/v1/constrainttemplate_types.go
```diff
@@ -34,7 +34,8 @@ type CRD struct {
 }
 
 type CRDSpec struct {
-	Names      Names       `json:"names,omitempty"`
+	Names Names `json:"names,omitempty"`
+	// +kubebuilder:default={legacySchema: false}
```
julianKatz, line 38: This strikes me as the case of `Validation` being a `nil` pointer, which then should be defaulted to:

``` { legacySchema: false }```

But that seems different than the test cases that were added in `constrainttemplate_types_test.go`, which are about `Validation.OpenAPIV3Schema` being `nil`.

Can you elaborate?

Comment (julianKatz) [reply constraint/pkg/apis/templates/v1/constrainttemplate_types.go:38]: I see now that the two defaults serve two different cases: `Validation: nil` and `Validation: &Validation{}`

File: constraint/deploy/crds.yaml (hunk 1 above)
Comment (julianKatz) [reply constraint/deploy/crds.yaml:146]: I see now that it serves two cases: `Validation: nil` and `Validation: &Validation{}`

File: constraint/pkg/apis/templates/v1beta1/conversion.go [hunk 3]
```diff
@@ -24,8 +24,14 @@ import (
 )
 
 func Convert_v1beta1_Validation_To_templates_Validation(in *Validation, out *coreTemplates.Validation, s conversion.Scope) error { //nolint:golint
-	if in.OpenAPIV3Schema != nil {
-		inSchemaCopy := in.OpenAPIV3Schema.DeepCopy()
+	inSchema := in.OpenAPIV3Schema
+	// legacySchema should allow for users to provide arbitrary parameters, regardless of whether the user specified them
+	if in.LegacySchema && inSchema == nil {
+		inSchema = &apiextensionsv1.JSONSchemaProps{}
```
julianKatz, line 30: This strikes me as a defaulting action, rather than a conversion action.  I think we'd be better off linking those two things together somehow.

constraint/pkg/apis/templates/v1/constrainttemplate_types.go
maxsmythe: If `validation` is nil, then K8s defaulting doesn't go further, which means the default value for `legacySchema` is never applied.

The test validates that `legacySchema` behaves properly when `openAPIV3Schema` is nil. The extra defaulting makes it so that that logic actually gets invoked.

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
```diff
@@ -147,6 +141,8 @@ spec:
                             type: array
                         type: object
                       validation:
+                        default:
+                          legacySchema: true
```
ritazh, line 145: Thanks for working on this PR @maxsmythe! With this change, users would still need to update their CRDs right?

constraint/pkg/apis/templates/v1beta1/conversion.go
maxsmythe: This is a conversion action.

In the case where `legacySchema == false`, the schema remains nil and the expectation is that no parameters are provided.

When `legacySchema == true`, users could provide any value to parameters and it would be accepted, this preserves that.

Essentially we are unpacking the implicit `x-preserves` embedded in `legacySchema`, no new intent is being injected (which defaulting would do)

julianKatz (approved):

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
maxsmythe: You're welcome! :)

Users still need to update their CRDs. In general, it seems dangerous to allow CRDs to go stale.

I talk a bit more about the CRD thing in Oren's issue:

https://github.com/open-policy-agent/gatekeeper/issues/1458

Basically saying that we could paper over having an old CRD, but that would eventually break anyway once we moved to defaulting to v1 CRDs.

maxsmythe: Also, if we do want to paper over having an out-of-date CRD manifest, we'd have to do that in the Gatekeeper repo.

sozercan (approved):
LGTM

PR #132, opened 2021-08-10
Migrate large files to lfs

willbeason:
This doesn't edit the repository history. If we instead want to edit repository history, we'll need to force push to master.


willbeason:
It appears this didn't actually work

maxsmythe:
It looks like there will be another way to avoid large files coming down the pipe:

https://github.com/open-policy-agent/opa/pull/3708

willbeason:
This path didn't work. See #133

PR #133, opened 2021-08-11
Disable downloading large files by default

willbeason:
Right now we're using LFS for large files, but by default when cloning
the repositories, callers ask for the real versions of large files. By
setting these settings in the .gitconfig file, we change the default to
not download these large files when cloning the repository.


sozercan (approved):

maxsmythe (approved):
LGTM

PR #134, opened 2021-08-18
add external data provider

sozercan:

Design doc: https://docs.google.com/document/d/1hPi86jdsCKg8puYT5_s_73mPGExUJeZfyKmvG-XWtPc/edit


constraint/pkg/client/drivers/local/local.go
```diff
@@ -75,22 +85,82 @@ func New(args ...Arg) drivers.Driver {
...
+				if err := ast.As(a.Value, &providerName); err != nil {
+					return nil, err
+				}
+				if err := ast.As(b.Value, &body); err != nil {
+					return nil, err
+				}
+
+				provider, err := d.providerCache.Get(providerName)
+				if err != nil {
+					return nil, fmt.Errorf("unable to retrieve provider %v cache", providerName)
```
maxsmythe, line 133: nit: "from cache"

constraint/pkg/client/drivers/local/local.go
```diff
@@ -75,22 +85,82 @@ func New(args ...Arg) drivers.Driver {
 	for _, arg := range args {
 		arg(d)
 	}
+
+	// adding externaldata builtin otherwise capabilities get overridden
+	// if a capability, like http.send, is disabled
+	if d.providerCache != nil {
+		d.capabilities.Builtins = append(d.capabilities.Builtins, &ast.Builtin{
+			Name: "externaldata",
+			Decl: opatypes.NewFunction(opatypes.Args(opatypes.S, opatypes.S), opatypes.A),
```
maxsmythe, line 94: It seems like requiring a string may be a bit limiting... what if a user wants to pass a more complex object?

Also... is there any way for a user to batch requests?

constraint/pkg/client/drivers/local/local.go
```diff
@@ -75,22 +85,82 @@ func New(args ...Arg) drivers.Driver {
...
+				if err := ast.As(b.Value, &body); err != nil {
+					return nil, err
+				}
+
+				provider, err := d.providerCache.Get(providerName)
+				if err != nil {
+					return nil, fmt.Errorf("unable to retrieve provider %v cache", providerName)
+				}
+
+				req, err := http.NewRequest("GET", provider.Spec.ProxyURL, bytes.NewBuffer([]byte(body)))
```
maxsmythe, line 136: It looks like we're just passing the value through wholesale. Do we want to define a protocol here so we can preserve the ability to pass things like additional data/context or do batching?

constraint/pkg/client/drivers/local/local.go
```diff
@@ -75,22 +85,82 @@ func New(args ...Arg) drivers.Driver {
...
+				if resp.StatusCode != http.StatusOK {
+					return nil, fmt.Errorf(resp.Status)
+				}
+
+				v, err := ast.ValueFromReader(resp.Body)
+				if err != nil {
+					return nil, err
+				}
+
+				return ast.NewTerm(v), nil
```
maxsmythe, line 160: It looks like we're foregoing local caching for now?

That's fine as long as we add before GA, IMO.

constraint/pkg/externaldata/cache.go
```diff
@@ -0,0 +1,51 @@
...
+		return v, nil
+	}
+	return v1alpha1.Provider{}, fmt.Errorf("key is not found in provider cache")
+}
+
+func (c *ProviderCache) Upsert(provider *v1alpha1.Provider) error {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
+	c.cache[provider.GetName()] = v1alpha1.Provider{
```
maxsmythe, line 32: use `provider.DeepCopy()` instead?

constraint/pkg/externaldata/cache.go
```diff
@@ -0,0 +1,51 @@
...
+			FailurePolicy: provider.Spec.FailurePolicy,
+			Timeout:       provider.Spec.Timeout,
+			MaxRetry:      provider.Spec.MaxRetry,
+		},
+	}
+
+	return nil
+}
+
+func (c *ProviderCache) Remove(name string) error {
```
maxsmythe, line 44: It doesn't look like an error is possible here, so we don't need to return an error.

constraint/pkg/externaldata/cache.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func NewCache() *ProviderCache {
+	return &ProviderCache{
+		cache: make(map[string]v1alpha1.Provider),
+	}
+}
+
+func (c *ProviderCache) Get(key string) (v1alpha1.Provider, error) {
+	if v, ok := c.cache[key]; ok {
```
maxsmythe, line 22: need to acquire a read lock

constraint/pkg/externaldata/cache.go
```diff
@@ -0,0 +1,51 @@
...
+		return v, nil
```
maxsmythe, line 23: Should we return a deepcopy to prevent callers from corrupting the cache?

constraint/pkg/externaldata/cache.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return v1alpha1.Provider{}, fmt.Errorf("key is not found in provider cache")
+}
+
+func (c *ProviderCache) Upsert(provider *v1alpha1.Provider) error {
```
maxsmythe, line 28: Do you foresee adding validation code? If so, returning an error here may make sense.

maxsmythe (changes requested):
Awesome! Mostly clear cut comments. The thorniest comments are around the Rego function signature and whether we should have a more well-defined wire format.

constraint/pkg/client/drivers/local/local.go
ritazh: Are you thinking additional data and batching could be part of the provider spec that can be evaluated here before making the http request?

maxsmythe: I'm thinking it's an important part of the requester/provider protocol we're establishing that we are missing since we aren't defining anything about what a request/response between Gatekeeper and the provider would look like.

A (kind of) related example: consider that K8s admission requests have structured request/response objects that allow us to write a webhook that is compatible across K8s versions and allows them to do things like add additional request data without breaking us and also allows them to interpret our results in a consistent manner. Since we are also creating a REST API, we should strive for the same kinds of backwards compatibility.

Two themes I'm thinking of here:

- We should define a stable protocol so that people can write providers and we can support them in a backward-compatible manner even if we introduce new features, such as (not saying all of these will necessarily happen, but they are common REST extensions):
  - Pagination
  - Batching
  - Auth (assuming bearer token)
  - Cache/resumption keys
  - Error reasons/ failure-mode hints (e.g. it's worth retrying, downstream problem, etc.), similar to how K8s added a "warning" severity to admission webhook responses.
  - Permissions delegation tokens
  - Common request context (e.g. requesting user)
- Having a well-defined place for parameters to be passed allows us to cache based off of the `(provider, parameter)` tuple. Caching is only possible if we know an answer will be consistent for a given request and we only know that if we can know where all the variables that could change a response are provided. Examples:
  - assuming we have configurable timeouts in the future, it should be possible to send a request with the same parameters but a different timeout and still get a cache hit.
  - In the case of different batches of requests, we should be able to unpack the keys being requested and see if they've already been cached. If I have requested `[1, 2, 3, 4]` before, and now I'm requesting `[1, 2]`, we should be able to serve that from the cache.

There may be more, but those are top-of-mind for me.

One example in the corresponding G8r PR:

```go
func SendProviderRequest(provider *externaldatav1alpha1.Provider, providerResponseCache map[types.ProviderCacheKey]string) (map[types.ProviderCacheKey]string, error) {

	body, err := json.Marshal(providerResponseCache)
 
	if err != nil {
		return nil, err
	}

	req, err := http.NewRequest("GET", provider.Spec.ProxyURL, bytes.NewBuffer(body))
	if err != nil {
		return nil, err
	}
	req.Header.Set("Content-Type", "application/json")
	timeout := time.Second * time.Duration(provider.Spec.Timeout)
	client := &http.Client{Timeout: timeout}
	resp, err := client.Do(req)
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()

	var respBody []byte
	if resp.StatusCode == 200 {
		respBody, err = ioutil.ReadAll(resp.Body)
		if err != nil {
			log.Error(err, "unable to read response body")
			return nil, err
		}
	}

	var result map[types.ProviderCacheKey]string
	err = json.Unmarshal(respBody, &result)
	if err != nil {
		return nil, err
	}

	return result, nil
}
```

We are exporting our cache directly to the provider, which implies that a provider should accept a format of `map[ProviderCacheKey]string` where `ProviderCacheKey` is defined as:

```go
// ProviderCacheKey is the map key for provider requests.
type ProviderCacheKey struct {
	ProviderName string `json:"providerName,omitempty"`
	// outbound data is based on Provider DataSource
	// it is either value of location or username in admission request
	OutboundData string     `json:"outboundData,omitempty"`
	DataSource   DataSource `json:"dataSource,omitempty"`
}
```

This means that:

- Rego code that wants to call the same provider must construct this same `map[ProviderCacheKey]string` request format (or else the Rego function needs to construct it), which pushes complexity on to Rego authors and could be problematic as `DataSource` only has meaning for mutation.
- We are coupling provider implementation (handled by an outside party) to our internal cache format. This means:
  - If we ever change the format of our cache, we risk breaking pre-existing providers
  - If we cannot break pre-existing providers, we are limited in how we change our cache format, which may limit our ability to improve performance or support different kinds of providers in the future
  - The request/response size grows with the size of the cache, which is bandwidth/latency/CPU inefficient (request size, deserialization cost), and leads to an overall N^2 complexity for total data sent per evaluation, where N is the number of mutators. A protocol that avoids exporting the cache would avoid this.
- As-written, we are creating a de-facto protocol that provider implementors (and Rego authors) need to be aware of, but it's not documented. By defining an explicit request/response interface, we can provide clarity and stability, and insulate Rego authors from the caching complexity.

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -0,0 +1,52 @@
...
+          kind:
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: ProviderSpec defines the desired state of Provider
+            properties:
+              failurePolicy:
+                type: string
```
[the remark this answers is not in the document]
sozercan: Thanks! Updated

constraint/config/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+resources:
+- crds/templates.gatekeeper.sh_constrainttemplates.yaml
+- crds/externaldata.gatekeeper.sh_providers.yaml
```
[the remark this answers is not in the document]
sozercan: added

constraint/pkg/client/client.go
```diff
@@ -74,6 +76,7 @@ type Client struct {
 	templates         map[templateKey]*templateEntry
 	constraints       map[schema.GroupKind]map[string]*unstructured.Unstructured
 	allowedDataFields []string
+	ProviderCache     externaldata.ProviderCache
```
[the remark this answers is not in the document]
sozercan: removed. weird that linter didn't gave a warning here.

constraint/pkg/client/drivers/local/local.go
sozercan: updated

constraint/pkg/externaldata/cache.go
sozercan: thanks! updated

constraint/pkg/externaldata/cache.go
sozercan: added

constraint/pkg/externaldata/cache.go
sozercan: updated

constraint/pkg/externaldata/cache.go
sozercan: removed

constraint/pkg/client/drivers/local/local.go
sozercan: I think we said we'll start with a string with this to keep things simple and consistent with mutation.

maxsmythe: I'm not sure that would be consistent with mutation as-written?

Mutation is allowing the user to pass the value of the field being mutated. That could be a:

- string
- struct
- list

basically any type.

If we want to try to keep things simple, that works for me, but whatever limitations we impose should be enforceable, forward-compatible, and consistent between mutation and validation.

WRT batching, I remember discussing accepting a list of requests for the argument, which would govern the batching behavior?

constraint/pkg/externaldata/cache.go
sozercan: added

constraint/pkg/apis/templates/yaml_constant.go
```diff
@@ -310,4 +310,58 @@ status:
     plural: ""
   conditions: []
   storedVersions: []
+---
```
julianKatz, line 313: This shouldn't be making its way in here.  It has the potential to break the logic in `pkg/apis/templates/crd_schema.go`.

Easiest fix I can think of is to have `controller-gen` put the CRDs each into their own files, so that we can grab the correct file in the Makefile.  (See the `CRD_SOURCE_FILE` variable in the [Makefile](https://github.com/open-policy-agent/frameworks/blob/master/constraint/Makefile#L66))

sozercan: looks like we were just copying crds.yaml to this. Changed it to `config/crds/templates.gatekeeper.sh_constrainttemplates.yaml`

sozercan: looks like there is a diff in preserveUnknownFields. Do you know why is there a difference between `crds.yaml` and `config/crds/templates.gatekeeper.sh_constrainttemplates.yaml`?

julianKatz: Is what you're referencing showing up in this PR?  I don't see `preserveUnknownFields` anywhere in this diff page.

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -0,0 +1,50 @@
...
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: ProviderSpec defines the desired state of Provider
```
maxsmythe, line 33: Can we add docstrings to the fields? Should be able to just put comments above the field names in go and they'll be created automatically

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -0,0 +1,50 @@
...
+            properties:
+              maxRetry:
```
maxsmythe, line 35: Maybe we should put max retries as part of the invocation of the provider? I could see validation more retry-tolerant than mutation, for example.

constraint/pkg/apis/templates/yaml_constant.go
```diff
@@ -16,7 +17,6 @@ spec:
     listKind: ConstraintTemplateList
     plural: constrainttemplates
     singular: constrainttemplate
-  preserveUnknownFields: false
```
maxsmythe, line 19: we probably want to keep the setting of `preserveUnknownFields` to `false`

constraint/pkg/client/drivers/local/local.go
```diff
@@ -75,22 +86,104 @@ func New(args ...Arg) drivers.Driver {
...
+			},
+			func(bctx rego.BuiltinContext, regorequest *ast.Term) (*ast.Term, error) {
+				var regoReq externaldata.RegoRequest
+				if err := ast.As(regorequest.Value, &regoReq); err != nil {
+					return nil, err
+				}
+				// only primitive types are allowed for keys
+				for _, key := range regoReq.Keys {
+					switch v := key.(type) {
+					case int:
```
maxsmythe, line 130: do we need to worry about `int32`, `int64`?

constraint/pkg/client/drivers/local/local.go
```diff
@@ -75,22 +86,104 @@ func New(args ...Arg) drivers.Driver {
...
+					case float32:
+						break
+					default:
+						return nil, fmt.Errorf("type %v is not supported in external_data", v)
+					}
+				}
+
+				provider, err := d.providerCache.Get(regoReq.ProviderName)
+				if err != nil {
+					return nil, fmt.Errorf("unable to retrieve provider %v from cache", regoReq.ProviderName)
```
maxsmythe, line 142: What happens when a builtin returns an error?

maxsmythe (commented):
Just a few questions, nothing major I don't think.

constraint/pkg/client/drivers/local/local.go
sozercan: added primitive type check for validation and rego should be able to batch requests using a list.

constraint/pkg/client/drivers/local/local.go
sozercan: added an API for ProviderRequest and ProviderResponse which is decoupled from the rego request and response types.

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
maxsmythe: @sozercan signal boost, I saw you commented on this PR since my review. Wanted to make sure you knew I was waiting for responses before I looked again

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
sozercan: you mean as part of `external_data()` (only for validation)? do we want to set up a default value if user doesn't provide this?

constraint/pkg/apis/templates/yaml_constant.go
sozercan: yea this is what i mentioned https://github.com/open-policy-agent/frameworks/pull/134#discussion_r694414463 here, checking why this is showing as diff

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
sozercan: thanks for the ping. missed your comments, added docstrings.

constraint/pkg/client/drivers/local/local.go
sozercan: added `int32` and `int64`. Omitted `bool` here, not sure if there would be a usecase for that.

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -0,0 +1,50 @@
...
+          metadata:
+            type: object
+          spec:
+            description: ProviderSpec defines the desired state of Provider
+            properties:
+              proxyURL:
+                description: ProxyURL is the URL of the proxy to use for the provider
+                type: string
+              timeout:
+                description: Timeout is the timeout for the provider
```
maxsmythe, line 39: nit: "when querying the provider"

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -0,0 +1,50 @@
...
+          kind:
+            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: ProviderSpec defines the desired state of Provider
+            properties:
+              proxyURL:
+                description: ProxyURL is the URL of the proxy to use for the provider
```
maxsmythe, line 36: Should we call this `proxyURL`? Since the provider itself is the proxy, this strikes me as potentially confusing, since people may think this is a proxy to the provider.

ValidatingAdmissionWebhooks just call this field `url`.

Also, should we add an example URL value so people don't need to guess if it needs an http prefix, etc.?

constraint/pkg/apis/templates/yaml_constant.go
maxsmythe: @julianKatz This is where we're losing `preserveUnknownFields`

constraint/pkg/client/drivers/local/local.go
maxsmythe: ack

constraint/pkg/client/drivers/local/local.go
maxsmythe: Does Rego panic or return the error to the user? Which behavior to we want?

sozercan: So I am talking to Torin about this. Looks like OPA changed behavior in 0.25:
https://github.com/open-policy-agent/opa/blob/main/CHANGELOG.md#0250

We can either set `rego.StrictBuiltinErrors(true)` or custom function can return `topdown.Halt`. However, looks like topdown.Halt is broken until v0.31 https://github.com/open-policy-agent/opa/issues/3534

constraint/pkg/externaldata/response.go
```diff
@@ -0,0 +1,51 @@
...
+// ProviderResponse is the API response from a provider.
+type ProviderResponse struct {
+	APIVersion string   `json:"apiVersion,omitempty"`
+	Kind       string   `json:"kind,omitempty"`
+	Response   Response `json:"response,omitempty"`
+}
+
+// Response is the struct that holds the response from a provider.
+type Response struct {
+	Idempotent  bool   `json:"idempotent,omitempty"`
```
maxsmythe, line 20: Should we comment on these fields as well?

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
ritazh: `url` sgtm

constraint/pkg/client/drivers/local/local.go
maxsmythe: Can we sidestep the whole issue for most errors?

If we consider "missing provider" and "error calling provider" as system errors and report them that way in `RegoResponse` accordingly, there will be no need to invoke OPA error behavior.

Other errors, like parsing errors, seem a bit more borderline, but I think we can still consider them "system errors"

ritazh: Do we treat all errors the same in terms of fail open vs fail close?

constraint/pkg/externaldata/cache.go
```diff
@@ -0,0 +1,74 @@
...
+		return false
+	}
+	if !strings.HasPrefix(url, "http://") && !strings.HasPrefix(url, "https://") {
+		return false
+	}
+	return true
+}
+
+func isValidTimeout(timeout int) bool {
+	return timeout >= 0
```
ritazh, line 73: Is there an opa timeout we want to use as an upper limit?

constraint/pkg/externaldata/request.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Request is the struct that contains the keys to query.
+type Request struct {
+	Keys []interface{} `json:"keys,omitempty"`
+}
+
+// NewRequest creates a new request for the external data provider.
+func NewProviderRequest(keys []interface{}) *ProviderRequest {
+	return &ProviderRequest{
+		APIVersion: "externaldata.gatekeeper.sh/v1",
```
ritazh, line 24: Do we want to start with `v1` here since this is the contract between Gatekeeper and providers? v1 signals stable?

constraint/pkg/client/drivers/local/local.go
sozercan: similarly, what happens if there is an error when sending to request back to provider?

sozercan: updated. returning `RegoResponse` for most errors

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
sozercan: updated

constraint/pkg/externaldata/request.go
sozercan: good catch, updated

constraint/pkg/externaldata/response.go
sozercan: added

constraint/pkg/apis/templates/yaml_constant.go
sozercan: do we still need this? looks like it might be deprecated in v1 CRD https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1/types.go#L68

constraint/pkg/client/drivers/local/local.go
sozercan: resolved since we are not going to cache in validation until we can do individual request cache

constraint/pkg/externaldata/request.go
maxsmythe: External data is an alpha feature, right? How do we know the API is stable?

sozercan:
@maxsmythe any unresolved comments?

constraint/pkg/externaldata/request.go
maxsmythe: But we're already returning `v1` here... are we saying the wire format is stable?

constraint/pkg/apis/externaldata/v1alpha1/provider_types.go
```diff
@@ -0,0 +1,58 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ProviderSpec defines the desired state of Provider.
+type ProviderSpec struct {
+	// URL is the URL of the proxy to use for the provider. URL is prefixed with http:// or https://.
+	URL string `json:"URL,omitempty"`
```
maxsmythe, line 25: In ValidatingAdmissionWebhookConfiguration it's `url` (lowercase)

constraint/pkg/apis/templates/yaml_constant.go
maxsmythe: We do still need it, otherwise existing users may still have `PreserveUnknownFields` set to `true`. It was actually added as part of the v1 CRD migration.

maxsmythe (commented):
Couple small things, but ready to merge after that.

constraint/pkg/externaldata/request.go
sozercan: That was a mistake, request is updated to be `v1alpha1`

ritazh: My comment above is saying we should not start with `v1`.

constraint/pkg/apis/templates/yaml_constant.go
sozercan: isn't it noop if it's set to true in v1? https://github.com/kubernetes/apiextensions-apiserver/blob/release-1.21/pkg/apis/apiextensions/validation/validation.go#L1442-L1445

sozercan: added steps to generate `spec.preserveUnknownFields`

constraint/pkg/externaldata/request.go
maxsmythe: Ah, oops, I read the comment the wrong way, sorry!

maxsmythe (approved):
LGTM!

ritazh (approved):

PR #136, opened 2021-08-20
update deprecated golint to revive

sozercan:

### DIFF
--- a/constraint/pkg/apis/templates/v1/conversion.go
+++ b/constraint/pkg/apis/templates/v1/conversion.go
@@ -23,7 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/conversion"
 )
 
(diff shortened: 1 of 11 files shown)


constraint/pkg/apis/templates/v1/conversion.go
```diff
@@ -23,7 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/conversion"
 )
 
-func Convert_v1_Validation_To_templates_Validation(in *Validation, out *coreTemplates.Validation, s conversion.Scope) error { //nolint:golint
```
willbeason, line 26: Are the nolints still required?

willbeason (approved):

constraint/pkg/apis/templates/v1/conversion.go
sozercan: yes because of underscores

PR #138, opened 2021-09-23
Move CRD defaulting helpers outside of pkg/apis

### LINKED ISSUE #137: ToVersionless should only call add to scheme once in a program's lifetime
Maybe use sync.Once?

julianKatz:
In the building of `gator` github.com/willbeason found that the usage of
a shared Scheme amongst multiple sub-packages of pkg/apis/templates was
not thread safe.  An internal map within the Scheme was experiencing a
simultaneous read and write.

Instantiating the scheme in a super-package and using it in the
sub-packages required the creation of a circular dependency, as the
super-package needed the contents of the sub-package to create a
functioning scheme that could convert between sub-package types.  This
was previously solved by adding sub-package contents to the scheme at
scheme-usage-time, but this yielded the race condition on the map.

This PR moves most of the core logic into a separate package: pkg/schema

To prevent a circular dependency, a *runtime.Scheme is passed from
consumer packages to this new package's function call.

These schemes are instantiated once in the consumer packages and the
output of pkg/schema is stored at consumer package init() time to
prevent unnecessary re-calculation.

Fixes #137


maxsmythe (approved):
LGTM

willbeason (approved):

sozercan (approved):
LGTM

PR #139, opened 2021-09-27
Upgrade frameworks version of golangci-lint

willbeason:
Use the same linters we use in gatekeeper/ for consistency.

Remove usage of deprecated/now-unsupported linters.

Make changes to make linting pass.

### DIFF
--- a/constraint/pkg/client/backend.go
+++ b/constraint/pkg/client/backend.go
@@ -42,7 +42,7 @@ func NewBackend(opts ...BackendOpt) (*Backend, error) {
 	return b, nil
 }
 
-// NewClient creates a new client for the supplied backend
+// NewClient creates a new client for the supplied backend.
 func (b *Backend) NewClient(opts ...Opt) (*Client, error) {
 	if b.hasClient {
 		return nil, errors.New("Currently only one client per backend is supported")
--- a/constraint/pkg/client/backend.go
+++ b/constraint/pkg/client/backend.go
@@ -2,10 +2,10 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers"
-	"errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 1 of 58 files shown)


constraint/pkg/client/backend.go
```diff
@@ -65,7 +66,7 @@ func (b *Backend) NewClient(opts ...Opt) (*Client, error) {
 	}
 	for _, field := range c.allowedDataFields {
 		if !validDataFields[field] {
-			return nil, errors.Errorf("Invalid data field %s", field)
+			return nil, fmt.Errorf("Invalid data field %s", field)
```
maxsmythe, line 69: nit: lowercase first letter?

constraint/pkg/client/drivers/local/local.go
```diff
@@ -107,20 +109,20 @@ func copyModules(modules map[string]*ast.Module, filter string) map[string]*ast.
 
 func (d *driver) checkModuleName(name string) error {
 	if name == "" {
-		return errors.Errorf("Module name cannot be empty")
+		return fmt.Errorf("Module name cannot be empty")
```
maxsmythe, line 112: nit here and below... lowercase first letter?

maxsmythe (approved):
LGTM with nits

sozercan (approved):
Thank you! LGTM pending @maxsmythe's comments

constraint/pkg/client/backend.go
willbeason: Done.

constraint/pkg/client/drivers/local/local.go
willbeason: Done. Fixed about a hundred such similar cases.

constraint/pkg/regorewriter/sources.go
```diff
@@ -91,18 +91,19 @@ func (s *Sources) Reparent(old, new string) error {
 
 // Write will write the sources to the filesystem.
 func (s *Sources) Write() error {
-	return s.forAll(func(module sourceFile) error {
+	// TODO: Determine if _ is intended to be used in anything.
```
sozercan, line 94: open TODO here

willbeason: I talked with Max about this - we weren't sure a great way to mark that this is obviously a bug in the code. It isn't worth the time to fix it and verify the original intent as it was written two years ago, is untested, and isn't used in any production code.

The intent is to leave a note for someone in the future (who presumably will care about the functionality of that code) that there is obviously a problem here, but we don't really know for sure what will fix it.

sozercan:
@willbeason good to merge?

PR #142, opened 2021-10-05
Fix scheme registration in pkg/apis/templates

julianKatz:
PR #138 introduced a bug: the type conversion functions required by the
Scheme for use in the ToVersionless functions of the v1, v1beta1,
v1alpha1 packages were missing.  This manifested itself as an error in
Gatekeeper:

"unable to convert template: converting (v1beta1.ConstraintTemplate) to
(templates.ConstraintTemplate): unknown conversion"

This problem was due to the ordering of init() functions.  init
functions are called according to the lexicographic order of their
containing files. zz_generated.conversion.go registers the conversion
functions with the scheme, but did so after the init() function
previously held in defaults.go due to the aforementioned ordering.  This
left the Scheme used in ToVersionless without the conversion functions
it was expected to have.

Rather than hack this ordering to ensure a correctly populated
`localSchemeBuilder`, this PR makes a schemeBuilder with the explicit
purpose of using it in the Scheme that's used in ToVersionless.  This
decouples the scheme from the ordering of init() funcs, resolving the
issue.

This PR also adds unit tests for each of the ToVersionless funcs,
ensuring such a problem won't happen again.

### DIFF
--- a/constraint/pkg/apis/templates/v1/scheme.go
+++ b/constraint/pkg/apis/templates/v1/scheme.go
@@ -1,0 +1,42 @@
+package v1
+
+import (
+	ctschema "github.com/open-policy-agent/frameworks/constraint/pkg/schema"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const version = "v1"
+
+var (
+	structuralSchema *schema.Structural
+	sch              *runtime.Scheme
+)
+
+func init() {
+	// Prevent problems with ordering of init() function calls.  These
+	// functions are called according to the lexicographic order of their
+	// containing files.  As Register() is called on the localSchemeBuilder by
+	// zz_generated.conversion.go, the conversion functions haven't been
+	// registered with the localSchemeBuilder by the time this init() function
+	// runs.  We sidestep this problem by adding RegisterConversions here.
+	schemeBuilder := runtime.NewSchemeBuilder(localSchemeBuilder...)
+	schemeBuilder.Register(RegisterConversions)
+
+	sch = runtime.NewScheme()
+	var err error
+	if err = apiextensionsv1.AddToScheme(sch); err != nil {
+		panic(err)
+	}
+	if err = apiextensions.AddToScheme(sch); err != nil {
+		panic(err)
+	}
+	if err = schemeBuilder.AddToScheme(sch); err != nil {
+		panic(err)
+	}
+	if structuralSchema, err = ctschema.CRDSchema(sch, version); err != nil {
+		panic(err)
+	}
+}
--- a/constraint/pkg/apis/templates/v1/defaults.go
+++ b/constraint/pkg/apis/templates/v1/defaults.go
@@ -1,37 +1,9 @@
 package v1
 
 import (
-	ctschema "github.com/open-policy-agent/frameworks/constraint/pkg/schema"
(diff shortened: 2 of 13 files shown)


sozercan (approved):
Thank you! LGTM

constraint/pkg/apis/templates/v1/scheme.go
```diff
@@ -0,0 +1,42 @@
...
+
+func init() {
+	// Prevent problems with ordering of init() function calls.  These
+	// functions are called according to the lexicographic order of their
+	// containing files.  As Register() is called on the localSchemeBuilder by
+	// zz_generated.conversion.go, the conversion functions haven't been
+	// registered with the localSchemeBuilder by the time this init() function
+	// runs.  We sidestep this problem by adding RegisterConversions here.
+	schemeBuilder := runtime.NewSchemeBuilder(localSchemeBuilder...)
+	schemeBuilder.Register(RegisterConversions)
```
maxsmythe, line 26: We should create a scheme builder that explicitly registers all necessary registrations. Otherwise this function will still be vulnerable to failures due to reordering that changes the contents of `localSchemeBuilder`

maxsmythe (commented):
LGTM after changes

constraint/pkg/apis/templates/v1/scheme.go
willbeason: Define the schemeBuilder outside of a function as a global var to avoid ordering concerns:

```
var schemeBuilder = runtime.NewSchemeBuilder(fn1, fn2, fn3, ...)
```

willbeason: It's okay to have our own package-internal scheme,
```
var toVersionlessSchemeBuilder = runtime.NewSchemeBuilder(...)
```

constraint/pkg/apis/templates/v1/scheme.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+func init() {
+	// Prevent problems with ordering of init() function calls.  These
+	// functions are called according to the lexicographic order of their
+	// containing files.  As Register() is called on the localSchemeBuilder by
+	// zz_generated.conversion.go, the conversion functions haven't been
+	// registered with the localSchemeBuilder by the time this init() function
+	// runs.  We sidestep this problem by adding RegisterConversions here.
+	sb := runtime.NewSchemeBuilder(SchemeBuilder.AddToScheme, addDefaultingFuncs)
```
julianKatz, line 25: @maxsmythe @willbeason 

While I got rid of `localSchemeBuilder...`, I'm still using `SchemeBuilder.AddToScheme`.  This is a package level variable created in [register.go](https://github.com/open-policy-agent/frameworks/blob/master/constraint/pkg/apis/templates/v1/register.go#L37).  I could replicate the code there within this `init()` function, or as package level variables in that file.  But, I didn't see a real upside to duplicating that code as the usage I have above already decouples me from other `init()` functions, which was the real problem.

LMK if you disagree.

constraint/pkg/apis/templates/v1/scheme.go
```diff
@@ -0,0 +1,42 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const version = "v1"
+
+var (
+	structuralSchema *schema.Structural
+	sch              *runtime.Scheme
```
willbeason, line 15: Use a more descriptive name than "sch". I'd recommend `versionedScheme`.

constraint/pkg/apis/templates/v1/scheme.go
willbeason: Looks good.

willbeason (approved):

PR #143, opened 2021-10-06
Skip broken test TestRemoteClientE2E

willbeason:
This test currently can't pass, so I've added t.Skip so that
"go test ./..." doesn't try to execute it. This makes running go tests
annoying.

Once the features it relies on are ready (and it is possible for the
test to pass) we can remove the skip.


willbeason:
Sent from wrong repository.

PR #146, opened 2021-10-13
Benchmark 3

willbeason:


willbeason:
Actual PR for benchmarks out for review.

PR #148, opened 2021-10-28
Benchmark 7

willbeason:
Example change that benchmarks moving Constraint matching logic to Go and implements sharding Rego runtime environments by ConstraintTemplate.


willbeason:
Actual PR for benchmarks out for review.

PR #150, opened 2021-11-11
Make Client E2E tests more flexible

willbeason:
Refactor tests so that:

- it is possible to specify rego and libraries per-test
- all tests use their own Client
- the Probe type is no longer necessary

This makes tests able to run hermetically. Additionally, In the future I'll be writing tests that use different Client configurations, which requires using different Clients for different tests (especially handler Rego).

Make response order more, but not completely, deterministic. We don't
need response order to be more deterministic for tests than what I've
put here, and as-is will make the output of reviewing objects more
consistent. There are common edge cases where ordering is still randomized,
but this can be fully fixed later.


willbeason:
GitHub isn't properly calculating the diffs. In all cases I've preserved the original test logic, and have only changed how the results are validated.

maxsmythe:
I wonder if there's some sort of non-determinism in the test coverage

constraint/pkg/client/e2e_test.go
```diff
@@ -0,0 +1,694 @@
...
+	denyTemplateWithLibRego = `package foo
+
+import data.lib.bar
+
+violation[{"msg": "DENIED", "details": {}}] {
+  bar.always[x]
+	x == "always"
+}`
+
+	denyTemplateWithLibLib = `package lib.bar
```
maxsmythe, line 87: Do we mean to use `LibLib` here?

maxsmythe (approved):
LGTM, 1 naming nit

constraint/pkg/client/e2e_test.go
willbeason: Yes. Otherwise you have "denyTemplateWithLibRego" and "denyTemplateWithLib" which isn't as clear. Really the names of the tests should be clearer, but since I'm not changing what is tested in this PR I don't want to change it.

constraint/pkg/client/e2e_test.go
```diff
@@ -0,0 +1,694 @@
...
+
+			if diff := cmp.Diff(want, got, cmpopts.IgnoreFields(types.Result{},
+				"Metadata", "Review", "Resource")); diff != "" {
+				t.Error(diff)
+			}
+		})
+	}
+}
+
+func TestE2EAutoreject(t *testing.T) {
```
julianKatz, line 262: How does `denied` compare to `rejection`?  Some comments above this test (and perhaps others) would be helpful for someone understanding the different usage patterns exercised by these various tests.

constraint/pkg/client/e2e_test.go
```diff
@@ -0,0 +1,694 @@
...
+			rsps, err := c.Review(ctx, targetData{Name: "Sara", ForConstraint: "Foo"})
+			if err != nil {
+				t.Fatalf("got Review: %v", err)
+			}
+
+			got := rsps.Results()
+			want := []*types.Result{{
+				Constraint:        cstr,
+				EnforcementAction: testEnforcementAction,
+				Msg:               "DRYRUN",
```
julianKatz, line 632: nit: might be nice to have this in the list of constants at the top of the file.  That way, a newcomer (me) would know that's one of the possible states when reading the test file

constraint/pkg/client/e2e_test.go
```diff
@@ -0,0 +1,694 @@
...
+}`
+
+	denyTemplateWithLibLib = `package lib.bar
+always[y] {
+  y = "always"
+}
+`
+)
+
+var denyAllCases = []struct {
```
julianKatz, line 94: Why not have positive cases as well?  I can imagine a scenario where a bug is introduced (perhaps in the rego library) that causes all templates be rejected.  These tests might still all/mostly pass.

julianKatz (approved):

constraint/pkg/client/e2e_test.go
willbeason: Good point. I'll add comments

constraint/pkg/client/e2e_test.go
willbeason: I agree. This PR is only for refactoring test cases so that adding such tests (which are indeed absolutely necessary) is easier. Having those in this PR would have made it too large.

constraint/pkg/client/e2e_test.go
willbeason: Done!

constraint/pkg/client/e2e_test.go
```diff
@@ -0,0 +1,705 @@
...
+			},
+		},
+	}
+}
+
+func newConstraint(kind, name string, params map[string]string, enforcementAction *string) *unstructured.Unstructured {
+	c := &unstructured.Unstructured{}
+	c.SetGroupVersionKind(k8schema.GroupVersionKind{
+		Group:   "constraints.gatekeeper.sh",
+		Version: "v1alpha1",
```
sozercan, line 56: do we want to update apiVersions?

ritazh (approved):

constraint/pkg/client/e2e_test.go
willbeason: Yes. There's actually a few changes I want to make related to string constants, but I'd like to have a PR just for that:

https://github.com/open-policy-agent/frameworks/issues/163

sozercan (approved):

PR #160, opened 2021-11-16
Go113errors

willbeason:
Move pkg/client to use go113 errors conventions.

In general, move to using constant base errors and wrap them with fmt.Errorf. Also make tests check that the expected error is returned.

In many cases I've extended tests to make them more flexible to accommodate this. I've made the tests more flexible, but haven't added examples of using that flexibility (for example, a client working with multiple targets). I intend to do so in a separate PR.

Fix a few concurrency bugs - the RW mutex wasn't properly guarding writes vs. reads for constraints. Also consider two Constraints with empty Specs to be "Semantically equal" as it is valid for Constraints to have empty specs.

### DIFF
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -32,13 +32,13 @@ type templateEntry struct {
 }
 
 type Client struct {
-	backend           *Backend
-	targets           map[string]TargetHandler
+	backend *Backend
+	targets map[string]TargetHandler
 
 	// mtx guards access to both templates and constraints.
-	mtx       sync.RWMutex
-	templates map[templateKey]*templateEntry
-	constraints       map[schema.GroupKind]map[string]*unstructured.Unstructured
+	mtx         sync.RWMutex
+	templates   map[templateKey]*templateEntry
+	constraints map[schema.GroupKind]map[string]*unstructured.Unstructured
 
 	allowedDataFields []string
 }
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -3,10 +3,9 @@ package client
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"path"
-	"regexp"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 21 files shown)


maxsmythe (approved):
Thanks for doing all this!

LGTM pending addressing the nil-check issue for SemanticEqual.

There is one other nit about when to generate the list of known targets.

Lastly, you may want to change the state of this PR away from "draft" if it's ready for review.

constraint/pkg/core/constraints/constraints.go
```diff
@@ -11,13 +11,7 @@ import (
...
 func SemanticEqual(c1 *unstructured.Unstructured, c2 *unstructured.Unstructured) bool {
-	s1, exists, err := unstructured.NestedMap(c1.Object, "spec")
-	if err != nil || !exists {
-		return false
-	}
-	s2, exists, err := unstructured.NestedMap(c2.Object, "spec")
-	if err != nil || !exists {
-		return false
-	}
+	s1 := c1.Object["spec"]
```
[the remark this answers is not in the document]
willbeason: Good idea - done! This also gives us a true short-circuit if the pointers are equal.

constraint/pkg/client/client.go
```diff
@@ -158,14 +122,18 @@ func (c *Client) validateTargets(templ *templates.ConstraintTemplate) (*template
...
 
-	if len(templ.Spec.Targets) != 1 {
-		return nil, nil, fmt.Errorf("expected exactly 1 item in targets, got %v", templ.Spec.Targets)
-	}
-
 	targetSpec := &templ.Spec.Targets[0]
 	targetHandler, found := c.targets[targetSpec.Target]
+
+	var knownTargets []string
+	for known := range c.targets {
```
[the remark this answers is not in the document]
willbeason: Done!

constraint/pkg/client/error_map.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		b.WriteString(fmt.Sprintf("%s: %s\n", k, (*e)[k]))
+	}
+	return b.String()
+}
+
+func (e *ErrorMap) Is(target error) bool {
```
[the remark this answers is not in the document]
willbeason: Acknowledged. It's possible as-is, but I wouldn't add it unless multiple callers need that functionality. Since ErrorMap is a non-struct non-interface type definition *(is there a shorter name for that?)* callers are able to do this themselves.

constraint/pkg/client/errors.go
```diff
@@ -58,3 +59,27 @@ func (e *MissingTemplateError) Error() string {
 func NewMissingTemplateError(mapKey string) error {
 	return &MissingTemplateError{mapKey}
 }
+
+// Errors is a list of error.
+//
+// Deprecated: Use a structured result type if it is important to disambiguate
```
julianKatz: Does this mean have individual custom types for each error?

constraint/pkg/core/constraints/constraints.go
```diff
@@ -11,13 +11,15 @@ import (
...
-		return false
+	if c1 == c2 {
+		return true
 	}
-	s2, exists, err := unstructured.NestedMap(c2.Object, "spec")
-	if err != nil || !exists {
+
+	if (c1 == nil) != (c2 == nil) {
 		return false
 	}
```
julianKatz, line 20: nit: I don't think this second clause needs to be so complicated.

It could just be:

```golang
if c1 == nil || c2 == nil {
  return false
}
```

The first clause will catch when both pointers are `nil` (as they will be equal), and so we only need to worry about a single one of them being nil.  See this [playground](https://goplay.space/#MEENPSQFTVp) where I tested that.

constraint/pkg/core/constraints/constraints_test.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+func TestSemanticEqual(t *testing.T) {
+	testCases := []struct {
+		name string
+		c1   *unstructured.Unstructured
+		c2   *unstructured.Unstructured
+		want bool
+	}{
+		{
```
julianKatz, line 16: nit: a testcase for passing the same pointer might also be useful, as that's covered by your logic

julianKatz (approved):

constraint/pkg/client/errors.go
willbeason: It's very situational. If we want to report individual error codes, then we should use a structured error type. If we need to programmatically work with information the error can communicate, then the error should have that structure. In general we want to treat errors as black-box as possible unless we have a functional reason to do otherwise.

constraint/pkg/core/constraints/constraints.go
willbeason: I found a more amusing solution that removes another if statement:
```
if c1 == nil || c2 == nil {
	return c1 == c2
}
```

Issue #157, opened 2021-11-15
Increase unit test coverage for constraint/pkg/regorewriter


willbeason:
Coverage is ~60% and there's a lot of code execution paths which look easy to test but don't have tests yet.

Add tests for these.

willbeason:
To test coverage:

`go test ./pkg/regorewriter --coverprofile=cover.out`

to see coverage line-by-line:

`go tool cover -html=cover.out`

becky-hd:
+1

willbeason:
Done!

PR #161, opened 2021-11-18
Add unit tests for regowriter

becky-hd:
Increase unit test coverage for constraint/pkg/regorewriter #157

### DIFF
--- a/constraint/pkg/regorewriter/errors.go
+++ b/constraint/pkg/regorewriter/errors.go
@@ -1,6 +1,7 @@
 package regorewriter
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"strings"
@@ -41,3 +42,11 @@ func (errs Errors) Format(s fmt.State, verb rune) {
 		_, _ = fmt.Fprintf(s, "%q", errs.Error())
 	}
 }
+
+var (
+	ErrInvalidModule  = errors.New("invalid module")
+	ErrInvalidImport  = errors.New("invalid import")
+	ErrInvalidLibs    = errors.New("invalid lib prefix")
+	ErrDataReferences = errors.New("invalid Data References")
+	ErrReadingFile    = errors.New("error reading file")
+)
--- a/constraint/pkg/regorewriter/errors.go
+++ b/constraint/pkg/regorewriter/errors.go
@@ -47,6 +47,6 @@ var (
 	ErrInvalidModule  = errors.New("invalid module")
 	ErrInvalidImport  = errors.New("invalid import")
 	ErrInvalidLibs    = errors.New("invalid lib prefix")
-	ErrDataReferences = errors.New("invalid Data References")
+	ErrDataReferences = errors.New("invalid data references")
 	ErrReadingFile    = errors.New("error reading file")
 )
--- a/constraint/pkg/regorewriter/regorewriter.go
+++ b/constraint/pkg/regorewriter/regorewriter.go
@@ -358,9 +358,6 @@ func (r *RegoRewriter) rewriteDataRef(ref ast.Ref) ast.Ref {
 // rewriteImportPath updates an import path to the new value.
 func (r *RegoRewriter) rewriteImportPath(path *ast.Term) error {
 	glog.V(vLogDetail).Infof("import: %s %#v", path, path)
-	for _, t := range path.Value.(ast.Ref) {
-		glog.V(vLogDetail).Infof("  term: %s %#v %#v", t, t, reflect.TypeOf(t.Value).String())
-	}
 	pathRef, ok := path.Value.(ast.Ref)
 	if !ok {
 		return fmt.Errorf("got reference of type %T, want %T", path.Value, ast.Ref{})
--- a/constraint/pkg/regorewriter/regorewriter.go
+++ b/constraint/pkg/regorewriter/regorewriter.go
@@ -40,7 +40,7 @@ type RegoRewriter struct {
 
 // New returns a new RegoRewriter
 // args:
-// 	it - the PackageTransformer that will be used for updating the path
+// 	pt - the PackageTransformer that will be used for updating the path
 //	libs - a list of package prefixes that are allowed for library use
 //	externs - a list of packages that the rego is allowed to reference but not declared in any libs
 func New(pt PackageTransformer, libs []string, externs []string) (*RegoRewriter, error) {
@@ -64,7 +64,7 @@ func New(pt PackageTransformer, libs []string, externs []string) (*RegoRewriter,
 func (r *RegoRewriter) add(path, src string, slice *[]*Module) error {
 	m, err := ast.ParseModule(path, src)
 	if err != nil {
-		return err
+		return fmt.Errorf("%w: %v", ErrInvalidModule, err)
 	}
 
 	r.addModule(path, m, slice)
@@ -95,7 +95,7 @@ func (r *RegoRewriter) addTestDir(testDirPath string) error {
 	glog.V(vLog).Infof("Walking test dir %s", testDirPath)
 	walkFn := func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			return fmt.Errorf("walk error on path %s: %w", path, err)
+			return fmt.Errorf("%w: walk error on path %s: %v", ErrReadingFile, path, err)
 		}
 		if info.IsDir() {
 			return nil
@@ -107,7 +107,7 @@ func (r *RegoRewriter) addTestDir(testDirPath string) error {
 		glog.V(vLog).Infof("reading %s", path)
 		bytes, err := ioutil.ReadFile(path)
 		if err != nil {
-			return err
+			return fmt.Errorf("%w: %v", ErrReadingFile, err)
 		}
 
 		r.testData = append(r.testData, &TestData{FilePath: FilePath{path: path}, content: bytes})
@@ -126,7 +126,7 @@ func (r *RegoRewriter) addFileFromFs(path string, slice *[]*Module) error {
 
 	bytes, err := ioutil.ReadFile(path)
 	if err != nil {
-		return err
+		return fmt.Errorf("%w: %v", ErrReadingFile, err)
 	}
 	return r.add(path, string(bytes), slice)
 }
@@ -141,13 +141,13 @@ func (r *RegoRewriter) addFileFromFs(path string, slice *[]*Module) error {
 func (r *RegoRewriter) addPathFromFs(path string, slice *[]*Module) error {
 	fileStat, err := os.Stat(path)
 	if err != nil {
-		return err
+		return fmt.Errorf("%w: %v", ErrReadingFile, err)
 	}
 
 	if fileStat.IsDir() {
 		infos, err := ioutil.ReadDir(path)
 		if err != nil {
-			return err
+			return fmt.Errorf("%w: %v", ErrReadingFile, err)
 		}
 
 		// handle test dirs
@@ -223,7 +223,7 @@ func (r *RegoRewriter) checkLibPackages() error {
 	for _, mod := range r.libs {
 		path := mod.Module.Package.Path
 		if !r.allowedLibPackage(path) {
-			return fmt.Errorf("path %s not found in lib prefixes", path)
+			return fmt.Errorf("%w: path %s not found in lib prefixes", ErrInvalidLibs, path)
 		}
 	}
 	return nil
@@ -275,11 +275,11 @@ func (r *RegoRewriter) checkImport(i *ast.Import) error {
 
 	importRef, ok := i.Path.Value.(ast.Ref)
 	if !ok {
-		return fmt.Errorf("got reference of type %T, want %T", i.Path.Value, ast.Ref{})
+		return fmt.Errorf("%w: got reference of type %T, want %T", ErrInvalidImport, i.Path.Value, ast.Ref{})
 	}
 
 	if isSubRef(inputRefPrefix, importRef) {
-		return fmt.Errorf("bad import")
+		return fmt.Errorf("%w: bad import", ErrInvalidImport)
 	}
 
 	for _, libPrefix := range r.allowedLibPrefixes {
@@ -288,7 +288,7 @@ func (r *RegoRewriter) checkImport(i *ast.Import) error {
 		}
 	}
 
-	return fmt.Errorf("bad import")
+	return fmt.Errorf("%w: bad import", ErrInvalidImport)
 }
 
 // checkDataReferences checks that all data references are directed to allowed lib prefixes or
@@ -312,7 +312,7 @@ func (r *RegoRewriter) checkDataReferences() error {
 			})
 		}
 		if errs != nil {
-			return fmt.Errorf("check refs failed on module %s: %w", m.FilePath, errs)
+			return fmt.Errorf("%w: check refs failed on module %s: %v", ErrDataReferences, m.FilePath, errs)
 		}
 		return nil
 	})
@@ -356,17 +356,17 @@ func (r *RegoRewriter) rewriteDataRef(ref ast.Ref) ast.Ref {
 }
 
 // rewriteImportPath updates an import path to the new value.
-func (r *RegoRewriter) rewriteImportPath(path *ast.Term) (*ast.Term, error) {
+func (r *RegoRewriter) rewriteImportPath(path *ast.Term) error {
+	glog.V(vLogDetail).Infof("import: %s %#v", path, path)
+	for _, t := range path.Value.(ast.Ref) {
+		glog.V(vLogDetail).Infof("  term: %s %#v %#v", t, t, reflect.TypeOf(t.Value).String())
+	}
 	pathRef, ok := path.Value.(ast.Ref)
 	if !ok {
-		return nil, fmt.Errorf("got reference of type %T, want %T", path.Value, ast.Ref{})
+		return fmt.Errorf("got reference of type %T, want %T", path.Value, ast.Ref{})
 	}
-
-	if !r.refNeedsRewrite(pathRef) {
-		return path, nil
-	}
-
-	return ast.NewTerm(r.rewriteDataRef(pathRef)), nil
+	path.Value = r.rewriteDataRef(pathRef)
+	return nil
 }
 
 // Rewrite will check the input source and update the package paths and refs as appropriate.
@@ -383,13 +383,7 @@ func (r *RegoRewriter) Rewrite() (*Sources, error) {
 	// libs, entryPoints - update import and other refs
 	err := r.forAllModules(func(mod *Module) error {
 		for _, i := range mod.Module.Imports {
-			glog.V(vLogDetail).Infof("import: %s %#v", i.Path, i.Path)
-			for _, t := range i.Path.Value.(ast.Ref) {
-				glog.V(vLogDetail).Infof("  term: %s %#v %#v", t, t, reflect.TypeOf(t.Value).String())
-			}
-			var err error
-			i.Path, err = r.rewriteImportPath(i.Path)
-			if err != nil {
+			if err := r.rewriteImportPath(i.Path); err != nil {
 				return err
 			}
 		}
--- a/constraint/pkg/regorewriter/regorewriter_test.go
+++ b/constraint/pkg/regorewriter/regorewriter_test.go
@@ -94,11 +94,10 @@ func RegoSrc(pkg string, opts ...RegoOption) string {
 
 // RegoRewriterTestcase is a testcase for rewriting rego.
 type RegoRewriterTestcase struct {
-	name       string            // testcase name
-	baseSrcs   map[string]string // entrypoint files
-	libSrcs    map[string]string // lib files
-	wantError  error             // error returned when RegoRewriter should reject input
-	wantResult map[string]string // expected output files from RegoRewriter
+	name      string            // testcase name
+	baseSrcs  map[string]string // entrypoint files
+	libSrcs   map[string]string // lib files
+	wantError error             // error returned when RegoRewriter should reject input
 }
 
 func MockRegoWriter() (*RegoRewriter, error) {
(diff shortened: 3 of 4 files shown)


maxsmythe:
It looks like the lint test is failing? Also, feel free to assign Will as the reviewer once this PR is ready for a look.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +499,56 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			wantError: true,
+		},
+		{
+			name:      "fail to add lib",
+			libSrcs:   map[string]string{"invalid.rego": RegoSrc("foo", Body("something invalid"))},
+			wantError: true,
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, tc.Run)
```
willbeason, line 517: I know this is the pattern in the file, but it obfuscates the functionality under test. As the test reader, from reading this test I can't tell what function is being called that returns the error.

Compare to:
```
tcs := []struct{
  name string
  path string
  src string
  wantError error
} {{
  name: "fail to add entrypoint",
  src: "something invalid",
}, ...}

for _, tc := range tcs {
  t.Run(tc.name, func() {
    rr, err := New(...)
    
    gotErr := rr.AddEntryPoint(tc.path, tc.content)
    if !errors.Is(tc.wantError, err) {
      t.Fatalf("got AddEntryPoint() error = %v, want %v", gotErr, tc.wantError)
    }
  })
}
```

This is fully self-contained. I don't have to go around the file to figure out what was being called, and what failed.

With that, as the test reader I can immediately see what is being tested at the exported API level rather than what is hidden away in a tc.Run() command. As a test writer, I can finely craft failure modes specific to the function being tested rather than worrying about making something that could break the ten or so other things being called in RegoRewriterTestcase.Run().

While this might result in more verbose test code, it is better to have more verbose (but clear) test code than to try to DRY it.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +499,56 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+	}
+}
+
+func TestRegoRewriterAddLibFromFs(t *testing.T) {
+	tests := []struct {
+		name    string
+		path    string
+		wantErr bool
```
willbeason, line 525: Create a base error for this failure mode with `errors.New()`. The function should wrap the base error with additional information, and then this test can test for the error with `errors.Is()`

This ensures we aren't just getting _some error_, but the particular error we're looking for.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +499,56 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			pp := NewPackagePrefixer("foo.bar")
+			libs := []string{"data.lib"}
+			externs := []string{"data.inventory"}
+			rr, err := New(pp, libs, externs)
+			if err != nil {
+				t.Fatalf("Failed to create %s", err)
```
willbeason, line 544: Mention the specific function which failed in the test failure message. For example:

```
t.Fatalf("got New() error = %v, want nil", err)
```

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +499,56 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			name:    "path not exist",
+			path:    "foo/bar",
+			wantErr: true,
+		},
+		{
+			name: "path without test",
+			path: "./",
+		},
+	}
+	for _, tt := range tests {
```
willbeason, line 537: Nit: prefer "tc" instead of "tt" for test cases. Feel free to disregard this comment.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +499,56 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			pp := NewPackagePrefixer("foo.bar")
```
willbeason, line 539: Are all three of these options required? It seems like a lot to specify for a simple test. If not, remove the ones that aren't required. If they are required, feel free to close this comment.

willbeason (changes requested):

constraint/pkg/regorewriter/regorewriter_test.go
willbeason: It's clearer now.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,133 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			rr, err := MockRegoWriter()
+			if err != nil {
+				t.Fatalf("Failed to create %s", err)
+			}
+			if gotErr := rr.AddEntryPoint(tc.path, tc.src); !errors.Is(gotErr, tc.wantError) {
+				t.Errorf("got AddEntryPoint() %s error = %v, want %v", tc.name, gotErr, tc.wantError)
```
willbeason, line 598: Don't put `tc.name` in failure messages. Since we're using `t.Error` and similar, the test case name is already in the failure block.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,133 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+	tcs := []struct {
+		name      string
+		path      string
+		src       string
+		wantError error
+	}{
+		{
+			name: "add lib.rego path",
+			path: "data.stuff.foolib",
+			src: RegoSrc("lib.rego", Body(`test_ok {
```
willbeason, line 579: Use the raw rego rather than helper methods for this (src should just be assigned a string rather than the result of a method) - as-is the test case hides what the literal rego source is.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,133 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			ref, err := ast.ParseRef(tt.imports)
+			if err != nil {
+				t.Fatalf("Failed to parse ref %s", err)
+			}
+			wantRef, err := ast.ParseRef(tt.want)
+			if err != nil {
+				t.Fatalf("Failed to parse ref %s", err)
+			}
+			if got := r.rewriteDataRef(ref); !got.Equal(wantRef) {
+				t.Errorf("rewriteDataRef() = %v, want %v", got, tt.want)
```
willbeason, line 563: Use %q instead of %v or %s for quoting string values to make error messages easier to read. This particularly helps if strings ever have whitespace or unusual characters.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,133 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+		{
+			name:    "add prefix same as libs",
+			prefix:  "data.lib",
+			imports: "data.lib.util",
+			want:    "data.data.lib.lib.util",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r, err := New(NewPackagePrefixer(tt.prefix), []string{"data.lib"}, []string{"data.ext", "data.inventory"})
```
willbeason, line 550: Right now `[]string{"data.lib"}` and `[]string{"data.ext", "data.inventory"}` are functioning as magic values - I don't know what they mean in the context of the test. Assign these to variables with descriptive names, like `allowedPrefix` so it's clearer what their purpose is.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -94,38 +97,45 @@ type RegoRewriterTestcase struct {
 	name       string            // testcase name
 	baseSrcs   map[string]string // entrypoint files
 	libSrcs    map[string]string // lib files
-	wantError  bool              // true if RegoRewriter should reject input
+	wantError  error             // error returned when RegoRewriter should reject input
 	wantResult map[string]string // expected output files from RegoRewriter
 }
 
-func (tc *RegoRewriterTestcase) Run(t *testing.T) {
+func MockRegoWriter() (*RegoRewriter, error) {
```
willbeason, line 104: It's better for tests to be explicit about the prefixer, libs, and externs they rely on. As-is the test cases give no hint that these values are special. It would be better to repeat the content of this method in each test so that (1) it is clear that these values have some special meaning in the tests and (2) it is easier to make the tests more flexible, e.g. if we want to test different values of `libs` or `externs`.

willbeason (changes requested):

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Ack. Removed tc.name, it was handy for searching failed tests.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+		{
+			name:    "add prefix same as libs",
+			prefix:  "data.lib",
+			imports: "data.some_ref",
+			want:    "data.data.lib.some_ref",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			externs := []string{"data.inventory"}
```
willbeason, line 553: Repeat this in each `extern` field for each test case. Defaulting makes tests confusing (e.g. what if I want to test behavior when there are no externs?)

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			if tt.extern != nil {
+				externs = tt.extern
+			}
+			r, err := New(NewPackagePrefixer(tt.prefix), []string{"data.lib"}, externs)
+			if err != nil {
+				t.Fatalf("Failed to create %s", err)
```
willbeason, line 559: Nit: It's always fine to just put `t.Fatal(err)`. If you find adding information to the error messages, make them explicit.  "Failed to create" doesn't say much. Instead, use a form like one of the below:

```
t.Fatalf("got New() error = %v, want nil", err)
```
or
```
t.Fatalf("failed to create RegoRewriter: %v", err)
```
The first example form is nice since it explicitly calls out the function which unexpectedly returned an error. The second is nice because it gives the context of what part of the test setup actually failed. It's also fine to combine the two, but it isn't necessary.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			want:    "data.data.lib.some_ref",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			externs := []string{"data.inventory"}
+			if tt.extern != nil {
+				externs = tt.extern
+			}
+			r, err := New(NewPackagePrefixer(tt.prefix), []string{"data.lib"}, externs)
```
willbeason, line 557: `"data.lib"` is a magic value here. It should either be a test case parameter (if we may want to test what happens when it is a different value) or assigned to a variable with a name that explains what it is.

We'll want to test different values of the libs argument (what if it's unset? does it change the behavior of the package prefixer?), we'll want this as an argument for the test case.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			ref, err := ast.ParseRef(tt.imports)
+			if err != nil {
+				t.Fatalf("Failed to parse ref %s", err)
+			}
+			wantRef, err := ast.ParseRef(tt.want)
+			if err != nil {
+				t.Fatalf("Failed to parse ref %s", err)
+			}
+			if got := r.rewriteDataRef(ref); !got.Equal(wantRef) {
+				t.Errorf("rewriteDataRef() = %q, want %v", got, tt.want)
```
willbeason, line 570: Nit: Since `got` and `want` are of the same type, either use `%q` or `%+v` for both of them so they're printed the same way.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			if tt.extern != nil {
+				externs = tt.extern
+			}
+			r, err := New(NewPackagePrefixer(tt.prefix), []string{"data.lib"}, externs)
+			if err != nil {
+				t.Fatalf("Failed to create %s", err)
+			}
+			ref, err := ast.ParseRef(tt.imports)
+			if err != nil {
+				t.Fatalf("Failed to parse ref %s", err)
```
willbeason, line 563: Include what we tried to parse in the error message. For example:

```
t.Fatalf("got ast.ParseRef(%q) error = %v, want nil", tt.imports, err)
```

(ditto others where we're parsing easily printable types like strings)

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+		content    string
+		wantResult map[string]string
+		wantError  error
+	}{
+		{
+			name:   "entry point imports lib",
+			prefix: "prefix.path",
+			content: `package lib.alpha
+				import data.lib.alpha
+        violation[{"msg":msg}] {
```
willbeason, line 664: The spacing for this is weird - a mix of spaces and tabs. Is that intentional and part of the test? If not, stick to tabs since `wantResult` uses tabs so it seems the rewriter normalizes to tabs.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			sources, err := rr.Rewrite()
+			if !errors.Is(err, tc.wantError) {
+				t.Errorf("Rewrite() got error = %v, want %v", err, tc.wantError)
+			}
+
+			result, err := sources.AsMap()
+			if err != nil {
+				t.Fatalf("unexpected error during Sources.AsMap: %s", err)
+			}
+			// TODO remove log
```
willbeason, line 753: (remove log when no longer necessary)

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			t.Logf("%v", result)
+
+			if diff := cmp.Diff(result, tc.wantResult); diff != "" {
```
willbeason, line 756: Standardize `cmp.Diff` to use want before got. So `cmp.Diff(tc.wantResult, result)`

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,256 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Failed to create %s", err)
+			}
+			if err := rr.AddEntryPoint("path", tc.content); err != nil {
+				if !errors.Is(err, tc.wantError) {
+					t.Fatalf("failed to add base source %q", err)
+				}
+				return
+			}
+			sources, err := rr.Rewrite()
```
willbeason, line 744: I'm confused - is this test for AddEntryPoint, or for Rewrite? Or do we have to Rewrite in order to observe the full effect of AddEntryPoint?

willbeason (changes requested):

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Removed default extern.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Updated the failed info for RegoRewriter creation failure.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Added lib to test parameter, also testing lib not provided cases.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Updated

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: removed

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Removed the indent in test source.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Swapped for want and got.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Renamed the test function to `RewriteEntryPoint` and `RewriteLib`, the tests are meant for rewrite the added entryPoint/lib.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Specified `libs` and `externs` are tested below in TestRegoRewriterRewriteDataRef, tested `prefix` in TestRegoRewriteLib.  Kept `MockRegoWriter` for test cases with fixed regorewriter initializer.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Updated.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Updated.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Keeping the original `TestRegoRewriter` that used the `tc.Run` function, but I rewrote the tests to `TestRegoRewriteEntryPoint` and `TestRegoRewriteLib`

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,369 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+	}
+}
+
+func TestRegoRewriteEntryPoint(t *testing.T) {
+	tests := []struct {
+		name       string
+		prefix     string
+		content    string
+		wantResult map[string]string
+		wantError  error
```
willbeason, line 657: Is `wantError` used by any tests? Either there should be a test that exercises it, or it should be removed.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,369 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Failed to create RegoRewriter %s", err)
+			}
+			if err := rr.AddLib("path", tc.content); err != nil {
+				if !errors.Is(err, tc.wantError) {
+					t.Fatalf("failed to add lib source %q", err)
+				}
+				return
+			}
+			sources, err := rr.Rewrite()
```
willbeason, line 859: Can `Rewrite()` ever fail? I don't see any tests that exercise this path. Either add a test that exercises that path, or if it can't reasonably fail then remove the handling logic.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,369 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+`},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			rr, err := New(NewPackagePrefixer(tc.prefix), []string{"data.lib"}, []string{"data.ext"})
+			if err != nil {
+				t.Fatalf("Failed to create RegoRewriter %s", err)
+			}
+			if err := rr.AddLib("path", tc.content); err != nil {
```
willbeason, line 853: AFAICT no tests expect AddLib to fail, so we don't need logic to check against `tc.wantError`, and we can just fail immediately.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -495,3 +506,369 @@ func TestRegoRewriterErrorCases(t *testing.T) {
...
+`},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			rr, err := New(NewPackagePrefixer(tc.prefix), []string{"data.lib"}, []string{"data.ext"})
+			if err != nil {
+				t.Fatalf("Failed to create RegoRewriter %s", err)
+			}
+			if err := rr.AddEntryPoint("path", tc.content); err != nil {
```
willbeason, line 738: No tests expect AddEntryPoint to fail, so we can remove the `errors.Is` check and can just immediately `t.Fatalf`

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -366,14 +376,14 @@ is_foo(name) {
 			libSrcs: map[string]string{
 				"lib/my_lib.rego": RegoSrc("lib"),
```
willbeason, line 377: Replace uses of `RegoSrc` with the strings it returns. As-is it's difficult to understand what the tests are doing.

willbeason (changes requested):

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Add this test case to `TestRegoRewriteLib`

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: replaced with 
```
if err != nil {
    t.Fatalf
}
```

constraint/pkg/regorewriter/regorewriter_test.go
willbeason: I'm confused. If `TestRegoRewriteLib` tests the same functionality, then we can delete the tests which use RegoSrc. In either case, we shouldn't have any uses of `RegoSrc` left.

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: removed `tc.wantError` here

willbeason (changes requested):

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: Added two Invalid lib test cases from `TestRegoRewriter`

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: removed

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: `TestRegoRewriteLib` and `TestRegoRewriteEntryPoint` separately cover AddLib/AddEntryPoint then Rewrite the source, used the same source content from `TestRegoRewriter`. We can probably either delete `TestRegoRewriter` or remove the template usage.

willbeason: Great! Then let's delete it now.

willbeason (approved):
Looks great, thanks Becky!

willbeason:
@ritazh @shomron ?

constraint/pkg/regorewriter/regorewriter.go
```diff
@@ -356,17 +356,17 @@ func (r *RegoRewriter) rewriteDataRef(ref ast.Ref) ast.Ref {
 }
 
 // rewriteImportPath updates an import path to the new value.
-func (r *RegoRewriter) rewriteImportPath(path *ast.Term) (*ast.Term, error) {
+func (r *RegoRewriter) rewriteImportPath(path *ast.Term) error {
+	glog.V(vLogDetail).Infof("import: %s %#v", path, path)
+	for _, t := range path.Value.(ast.Ref) {
```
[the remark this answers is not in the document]
becky-hd: removed.

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -1,11 +1,14 @@
...
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
 	"text/template"
 
+	"github.com/open-policy-agent/opa/ast"
+
```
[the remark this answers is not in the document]
becky-hd: sorted the import

julianKatz (approved):

maxsmythe (approved):
LGTM

constraint/pkg/regorewriter/errors.go
```diff
@@ -41,3 +42,11 @@ func (errs Errors) Format(s fmt.State, verb rune) {
 		_, _ = fmt.Fprintf(s, "%q", errs.Error())
 	}
 }
+
+var (
+	ErrInvalidModule  = errors.New("invalid module")
+	ErrInvalidImport  = errors.New("invalid import")
+	ErrInvalidLibs    = errors.New("invalid lib prefix")
+	ErrDataReferences = errors.New("invalid Data References")
```
sozercan, line 50: nit:
```suggestion
	ErrDataReferences = errors.New("invalid data references")
```

constraint/pkg/regorewriter/regorewriter_test.go
```diff
@@ -91,374 +93,128 @@ func RegoSrc(pkg string, opts ...RegoOption) string {
...
+			imports:   "data.stuff.foolib",
+			snippet:   "foolib.check(input)",
+			wantError: ErrInvalidImport,
 		},
 		{
 			name: "invalid binding of data to var",
 			snippet: `
-  x := data
-  x.stuff.more.stuff
+ x := data
```
sozercan, line 162: nit: can we align the snippet formatting? looks like some start with 2 spaces some with tabs

sozercan (commented):
minor nits, otherwise LGTM. 

Looks like GK tests are failing, is this expected?

constraint/pkg/regorewriter/regorewriter_test.go
becky-hd: replaced leading spaces with tabs

constraint/pkg/regorewriter/errors.go
becky-hd: Updated to lowercases.

willbeason:
> minor nits, otherwise LGTM.
> 
> @becky-hd Looks like GK tests are failing, is this expected?

Correct - this is currently a transient issue because we changed where validation is done. We're adding a dedicated "ValidateConstraintTemplate" method instead of having callers rely on implementation details of "CreateCRD".

sozercan (approved):
Thank you! LGTM

PR #162, opened 2021-11-19
Provide IsUnrecognizedConstraintError

willbeason:
For decoupling frameworks and gatekeeper error handling
(which allows using go113 error conventinos in tests), I need to remove
the dependence of gatekeeper on error types defined in frameworks so
these may be changed safely to raw errors (where reasonable to do so).

As an intermediate, I'm adding this function so that gatekeeper can
continue to compile, and the implementation of this method may be
switched out in frameworks.


willbeason:
This is safe since gatekeeper doesn't actually use the type; it just performs a type check.

maxsmythe (approved):
LGTM

sozercan (approved):
LGTM

PR #165, opened 2021-12-01
Use context consistently

willbeason:
Eliminate all uses of context.TODO(). Eliminate all uses of
context.Background() in production code. In all cases, either our
callers have a Context they can pass or the k8s framework provides one.

Remove unnecessary uses of context.Context. In quite a few places we
require context.Context, but nothing actually consumes it so it's just an
unused parameter that pollutes call sites and gives the false impression
that these requests are cancellable/etc. I've removed usages of Context
in these cases.

This change breaks gatekeeper - since this modifies our external APIs,
making this in a non-breaking way would require 5 PRs instead of 2. Once
this PR has been submitted, I'll submit a PR which updates the version
of frameworks/ that gatekeeper uses and fix the resulting breakages.

### DIFF
--- a/constraint/pkg/client/backend.go
+++ b/constraint/pkg/client/backend.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers"
@@ -23,8 +22,10 @@ func Driver(d drivers.Driver) BackendOpt {
 	}
 }
 
-// NewBackend creates a new backend. A backend could be a connection to a remote server or
-// a new local OPA instance.
+// NewBackend creates a new backend. A backend could be a connection to a remote
+// server or a new local OPA instance.
+//
+// A BackendOpt setting driver, such as Driver() must be passed.
 func NewBackend(opts ...BackendOpt) (*Backend, error) {
 	helper, err := newCRDHelper()
 	if err != nil {
@@ -36,7 +37,7 @@ func NewBackend(opts ...BackendOpt) (*Backend, error) {
 	}
 
 	if b.driver == nil {
-		return nil, errors.New("no driver supplied to the backend")
+		return nil, fmt.Errorf("%w: no driver supplied", ErrCreatingBackend)
 	}
 
 	return b, nil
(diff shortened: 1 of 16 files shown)


constraint/pkg/client/client.go
```diff
@@ -715,7 +715,7 @@ func (c *Client) init() error {
 		if err != nil {
 			return fmt.Errorf("could not re-format Rego source: %v", err)
 		}
-		if err := c.backend.driver.PutModule(context.Background(), modulePath, string(src)); err != nil {
+		if err := c.backend.driver.PutModule(ctx, modulePath, string(src)); err != nil {
```
maxsmythe, line 718: What happens if the context is interrupted mid put-module?

constraint/pkg/client/backend.go
```diff
@@ -44,7 +44,7 @@ func NewBackend(opts ...BackendOpt) (*Backend, error) {
 }
 
 // NewClient creates a new client for the supplied backend.
-func (b *Backend) NewClient(opts ...Opt) (*Client, error) {
+func (b *Backend) NewClient(ctx context.Context, opts ...Opt) (*Client, error) {
```
maxsmythe, line 47: Is NewClient a place we want to inject context?

willbeason: Like above, we don't actually use the context for anything so it just pollutes call sites. I've removed it.

constraint/pkg/client/client.go
willbeason: As it turns out, we don't need that Context. The issue Davis is working on will completely remove that, so I've just added a context.TODO() in `altermodules`.

julianKatz:
Are we confident that this decision can grow with the Constraint Framework?  What if we wanted to support another backend?  Perhaps a different implementation there would benefit from the ability to cancel requests.

No argument on the lack of utility at the moment, but we should be mindful of changing the public interface.  Other projects (like config validator) use the constraint framework as well, and we don't want them to break unless it's for a great reason.

maxsmythe:
We'll need to change the API as part of sharding Rego evaluation, so WRT backwards compatibility, as long as all these changes go out at once, it's probably not that much more pain.

As far as for future usage, it's a fair concern. IMO we should be able to inject context into requests that can/should be interrupted b/c they can have the potential to run for a long time. `Audit()` is a good example of this. Easy to cancel (no side effects), and long-running and resource-intensive enough to be worth canceling.

maxsmythe (approved):
LGTM

willbeason:
> Are we confident that this decision can grow with the Constraint Framework? What if we wanted to support another backend? Perhaps a different implementation there would benefit from the ability to cancel requests.
> 
> No argument on the lack of utility at the moment, but we should be mindful of changing the public interface. Other projects (like config validator) use the constraint framework as well, and we don't want them to break unless it's for a great reason.

Right now I'm tending towards YAGNI - since gatekeeper is the primary consumer, if we end up needing to add context back, I'd rather have to do it in the future and only add it where it's needed, than request it everywhere and just have it in places we end up never needing it.

julianKatz (approved):

Issue #171, opened 2021-12-16
[local.Driver] Replace PutModules and DeleteModules with AddTemplate and RemoveTemplate


willbeason:
```
func (d *Driver) AddTemplate(ct *templates.ConstraintTemplate) error
func (d *Driver) RemoveTemplate(ctx context.Context, ct *template.ConstraintTemplate) error
```

The driver itself should handle extracting libraries from the ConstraintTemplate and naming them - there's no need for Client to have any of this logic. Having Client not care how code from ConstraintTemplates is compiled is essential for us being able to shard ConstraintTemplates into their own environments under the hood.

For RemoveTemplate - note the inclusion of context.Context since this method will later be responsible for removing the corresponding Constraints.

It's fine to this and #170 as two separate PRs or as one PR - whatever works better for you.

becky-hd:
+1

willbeason:
Done!

PR #179, opened 2022-01-06
Add driver.AddTemplate and driver.RemoveTemplate

becky-hd:
#171 
Allow client.AddTemplate calling `driver.AddTemplate(*templates.ConstraintTemplate)` instead of parsing rego modules and calling `driver.PutModules` in client. 
client.CreateCRD uses the same parsing logic as part of template validation.


constraint/pkg/client/drivers/local/local.go
```diff
@@ -492,3 +497,192 @@ func (d *driver) Dump(ctx context.Context) (string, error) {
 
 	return string(b), nil
 }
+
+func MapModules(templ *templates.ConstraintTemplate, extern []string) (string, []string, error) {
```
willbeason, line 501: Add a comment which describes the intent of "MapModules". It is not clear what the intent is in general, or what "externs" mean for the function.

constraint/pkg/client/client.go
```diff
@@ -253,96 +240,20 @@ func (c *Client) createBasicTemplateArtifacts(templ *templates.ConstraintTemplat
...
 		return nil, fmt.Errorf("%w: got nil ConstraintTemplate",
 			ErrInvalidConstraintTemplate)
 	}
 
-	artifacts, err := c.createTemplateArtifacts(templ)
+	artifacts, err := c.createBasicTemplateArtifacts(templ)
 	if err != nil {
 		return nil, err
 	}
+	if _, _, err = local.MapModules(templ, c.allowedDataFields); err != nil {
```
willbeason, line 254: MapModules doesn't need to be called here. "MapModules" is already called from within `local.Driver.AddTemplate`, so we know we'll properly fail there. We aren't using any data from this call, so it can be safely removed.

constraint/pkg/client/client.go
```diff
@@ -363,18 +274,20 @@ func (c *Client) AddTemplate(templ *templates.ConstraintTemplate) (*types.Respon
...
-	if err != nil {
-		return resp, err
-	}
-
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
-	if err = c.backend.driver.PutModules(artifacts.namePrefix, artifacts.modules); err != nil {
-		return resp, fmt.Errorf("%w: %v", local.ErrCompile, err)
+	if d, ok := c.backend.driver.(interface{ AddExterns([]string) }); ok {
```
willbeason, line 280: When moving this to client.Init (see other review comments), instead cast to `local.Driver` and call this method. As-is we will not have a signal that this needs to be removed once AddExterns is removed from `local.Driver` (there won't be a compilation error to signal us that this is stale).

constraint/pkg/client/drivers/interface.go
```diff
@@ -29,11 +30,12 @@ type Driver interface {
 	// count of modules deleted.  Deletion of non-existing prefix will
 	// result in 0, nil being returned.
 	DeleteModules(namePrefix string) (int, error)
```
willbeason, line 32: Make sure to:
- [ ] Make "PutModules" and "DeleteModules" package-private since we don't need them any more. We don't want anyone else adding calls to them.
- [ ] Modify the tests for "PutModules" and "DeleteModules" to use "AddTemplate" and "DeleteTemplate" instead.

constraint/pkg/client/drivers/local/local.go
willbeason: Per comment in client.go, this can be made package-private.

constraint/pkg/client/drivers/local/local.go
```diff
@@ -492,3 +497,192 @@ func (d *driver) Dump(ctx context.Context) (string, error) {
...
+		return "", nil, err
+	}
+	targetSpec := templ.Spec.Targets[0]
+	targetHandler := targetSpec.Target
+	kind := templ.Spec.CRD.Spec.Names.Kind
+	libPrefix := templateLibPrefix(targetHandler, kind)
+
+	rr, err := regorewriter.New(
+		regorewriter.NewPackagePrefixer(libPrefix),
+		[]string{"data.lib"},
```
willbeason, line 512: Use a string constant for "data.lib" right now it's just a magic string. A good name for this would be "libRoot" with a comment saying that it is the root of all libraries a ConstraintTemplate declares.

willbeason (changes requested):

constraint/pkg/client/client.go
becky-hd: We would have to export local.driver to allow the type cast, I can add error for the !ok  when casts the SetExterns

willbeason: I see two good ways to resolve this - feel free to use either. My preference is 1, but I don't have a strong opinion.

1. It should be fine to export `local.driver` (arguably, it should be exported). All of its members are private so there's no danger of callers messing with its fields.

2. Alternatively, define an interface in `local/local.go` which defines an interface with `AddExterns` as its only method. Add a line `var _ AddExternser = &driver{}` which will fail to compile when we remove the method from `local.go`. `client.go` should cast to this type to get the method.

Either of these solutions would work.

willbeason (changes requested):

constraint/pkg/client/drivers/interface.go
```diff
@@ -3,6 +3,8 @@ package drivers
 import (
 	"context"
 
+	"github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
+
```
willbeason, line 7: remove empty line between these two packages

constraint/pkg/client/drivers/local/local.go
```diff
@@ -185,8 +192,8 @@ func (d *driver) PutModule(name string, src string) error {
 	return err
 }
 
-// PutModules implements drivers.Driver.
-func (d *driver) PutModules(namePrefix string, srcs []string) error {
+// PutModules upserts a number of modules under a given prefix.
```
willbeason, line 195: s/PutModules/putModules to reflect the name change.

willbeason (approved):

constraint/pkg/client/drivers/local/local.go
```diff
@@ -492,3 +497,192 @@ func (d *driver) Dump(ctx context.Context) (string, error) {
...
+			ErrInvalidConstraintTemplate)
+	case 1:
+		return nil
+	default:
+		return fmt.Errorf("%w: multi-target templates are not currently supported",
+			ErrInvalidConstraintTemplate)
+	}
+}
+
+func (d *driver) AddExterns(fields []string) {
```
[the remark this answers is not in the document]
becky-hd: renamed to `SetExterns` and moved to client.init

constraint/pkg/client/drivers/local/local.go
becky-hd: Moved back to `driver.AddTemplate`, we might need a validation function to resolve module parsing/validation in CreateCRD in the immediate future.

constraint/pkg/client/drivers/interface.go
becky-hd: trimed

constraint/pkg/client/client.go
becky-hd: Removed rego validation from CreateCRD, renamed `MapModules` to `ValidateConstraintTemplate`

sozercan:
@becky-hd looks like Gatekeeper tests are failing, is this expected?

becky-hd:
> @becky-hd looks like Gatekeeper tests are failing, is this expected?

Yes, this PR removed validation in CreateCRD results the some gatekeeper unit tests failure. I opened https://github.com/open-policy-agent/gatekeeper/issues/1788 as an immediate followup

ritazh (approved):
lgtm

PR #180, opened 2022-01-10
Add support for print built-in

anderseknert:
While playing around with the Gator CLI today I noticed
that I couldn't use the new print built-in from OPA in that
context. This function is really useful to quickly be able to
debug values of rules and variables, and would be a nice
addition to the Gator CLI (and possibly to the Gatekeeper
server too, but others can decide that). This PR adds options
to the local driver for enabling print and for providing a
printHook that decides where the output of print calls should
go. Will submit another PR for support in Gator if/when this
is merged.

Signed-off-by: Anders Eknert <anders@eknert.com>

### DIFF
--- a/constraint/pkg/client/e2e_test.go
+++ b/constraint/pkg/client/e2e_test.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"encoding/json"
+	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


srenatus (approved):
LGTM, one nitpick in the tests.

anderseknert:
@srenatus did you miss to post that nitpick? I can't find it 😄

constraint/pkg/client/e2e_test.go
```diff
@@ -463,6 +465,93 @@ func TestE2ERemoveTemplate(t *testing.T) {
...
+
+			results := rsps.Results()
+			if len(results) != 1 {
+				t.Errorf("expected 1 result, got %v", len(results))
+			}
+			if results[0].Msg != tc.wantMsg {
+				t.Errorf("expected msg %v, got %v", tc.wantMsg, results[0].Msg)
+			}
+
+			if len(tc.wantPrint) != 0 && len(printed) != 0 {
```
srenatus, line 546: My previously lost comment: if `len(tc.wantPrint) != 0` but `len(printed) == 0`, would the test fail?

anderseknert: Thanks! I've updated the logic to only run the deepEquals test if at least one of them has an element. Should work, huh?

srenatus (approved):

PR #181, opened 2022-01-10
Rewrite test TargetHandler

willbeason:
The big idea here is that the current TargetHandler makes it hard to test several things we want to have good tests for before we can safely shard rego environments. At its core, this PR is about this new handler and the resulting test changes. This makes `pkg/client/clienttest/handler.go` a good starting point for reviewing this PR.

I've broken "CRD Helpers" into individual methods and removed "CRDHelper" - this type was becoming a dumping ground of tenuously-related validation and constructing CRDs, ConstraintTemplates, and Constraints. Only one method used the single member of this type (the schema). So I've broken this into the `client/crds` package. The package could probably be better-named since not all of the functions in "CRDHelper" dealt with CRDs. Let me know if you have a better idea!

Since the test Handler needs to import the `client` package and is in a test package, all `client` tests need to be in the `client_test` package. All functionality they rely on then needs to be exported by `client`. This change is responsible for much of the code changes.

Lastly, the new test TargetHandler is much more flexible than the old one. This ended up making a lot of logic for existing E2E tests obsolete, so I've rewritten them. This makes the distinction between "e2e" and non-"e2e" tests a bit arbitrary as both flavors are using the same setup code (and have the same flexibility). I've arbitrarily decided to stick with the current convention that "e2e" tests are those which exercise "Review" or "Audit", or test sequences of behaviors (such as adding a Constraint, then deleting its Template).

Some changes in this PR are transitory - they are intended to be fixed in later PRs. These relate to making certain fields/methods exported which shouldn't be. `AllowedDataFields`, `ParseModule`, and `RequireModuleRules` are examples of these. They will be moved to `client/drivers/local` since that's the only location which should need that logic.


constraint/pkg/client/e2e_test.go
```diff
@@ -1,465 +1,453 @@
...
+		want        []*types.Result
+	}{
+		{
+			name:        "empty client returns empty audit",
+			templates:   nil,
+			constraints: nil,
+			objects:     nil,
+		},
+		{
+			name: "no template returns empty audit",
```
maxsmythe, line 347: nit: "no constraint returns empty audit"?

maxsmythe (approved):
LGTM with a few nits

constraint/pkg/client/e2e_test.go
willbeason: Good catch. Also added case for empty templates but with object.

constraint/pkg/client/clienttest/handler.go
```diff
@@ -0,0 +1,146 @@
...
+func (h *Handler) GetName() string {
+	return HandlerName
+}
+
+var libTempl = template.Must(template.New("library").Parse(`
+package foo
+
+autoreject_review[rejection] {
+	constraint := {{.ConstraintsRoot}}[_][_]
+  constraint.spec.autoreject
```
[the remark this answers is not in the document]
willbeason: Changed to only use spaces for indentation.

constraint/pkg/client/client_test.go
```diff
@@ -771,8 +737,24 @@ func TestTemplateCascadingDelete(t *testing.T) {
 		t.Error("could not remove template")
 	}
 
-	if len(c.constraints) != 1 {
-		t.Errorf("constraint cache expected to have only 1 entry: %+v", c.constraints)
+	_, err = c.GetConstraint(cst1)
+	if !errors.Is(err, client.ErrMissingConstraint) {
+		t.Errorf("found constraint %v %v", cst1.GroupVersionKind(), cst1.GetName())
```
ritazh, line 742: why do we not want log the err here?

constraint/pkg/client/client_test.go
```diff
@@ -771,8 +737,24 @@ func TestTemplateCascadingDelete(t *testing.T) {
...
+	}
+
+	_, err = c.GetConstraint(cst2)
+	if !errors.Is(err, client.ErrMissingConstraint) {
+		t.Errorf("found constraint %v %v", cst2.GroupVersionKind(), cst2.GetName())
```
ritazh, line 747: ditto

constraint/pkg/client/clienttest/templates.go
```diff
@@ -0,0 +1,180 @@
...
+		Rego:   moduleImportDenyRego,
+		Libs:   []string{moduleImportDenyLib},
+	}}
+
+	return ct
+}
+
+// moduleCheckData defines a Rego package which checks the "data" field of
+// Objects under review.
+// TODO: Also test "details" wiring.
```
ritazh, line 150: Is this to be addressed in another PR?

ritazh:
Gatekeeper Test  is failing

constraint/pkg/client/client_test.go
willbeason: Good catch! Done.

constraint/pkg/client/clienttest/templates.go
willbeason: Ah, I have already added those tests and forgot to remove the TODO.

constraint/pkg/client/client_test.go
willbeason: Good catch! Done.

willbeason:
> Gatekeeper Test is failing

Yes, this is expected for right now. @becky-hd is working on changing where/when we do validation. Right now Gatekeeper relies on CreateCRD to fail under certain conditions, when really we should have a dedicated ValidateCRD method.

ritazh (approved):
lgtm

Issue #170, opened 2021-12-16
[local.Driver] Add methods AddConstraint and RemoveConstraint


willbeason:
Use these signatures:

```
func (d *Driver) AddConstraint(ctx context.Context, target string, constraint *unstructured.Unstructured) error

func (d *Driver) RemoveConstraint(ctx context.Context, target string, constraint *unstructured.Unstructured) error
```

Note that relpath is not specified by Client - this logic needs to move to Driver as Client should not care about how Driver is organizing Constraints. So `createConstraintPath()` logic should be moved to the Driver. Note that `AddConstraint` should probably still call `PutData` for the required database transaction. And similar for `RemoveConstraint` - which should use `DeleteData`.

willbeason:
Done!

PR #182, opened 2022-01-14
Provide driver AddConstraint and RemoveConstraint (#170)

becky-hd:
Signed-off-by: Becky Huang <beckyhd@google.com>

### DIFF
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -37,8 +37,9 @@ type Client struct {
 	targets map[string]handler.TargetHandler
 
 	// mtx guards access to both templates and constraints.
-	mtx         sync.RWMutex
-	templates   map[templateKey]*templateEntry
+	mtx       sync.RWMutex
+	templates map[templateKey]*templateEntry
+	// TODO: https://github.com/open-policy-agent/frameworks/issues/187
 	constraints map[schema.GroupKind]map[string]*unstructured.Unstructured
 
 	AllowedDataFields []string
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -414,15 +414,6 @@ func createConstraintGKPath(target string, gk schema.GroupKind) string {
 	return constraintPathMerge(target, createConstraintGKSubPath(gk))
 }
 
-// createConstraintPath returns the storage path for a given constraint: constraints.<target>.cluster.<group>.<kind>.<name>.
-func createConstraintPath(target string, constraint *unstructured.Unstructured) (string, error) {
-	p, err := createConstraintSubPath(constraint)
-	if err != nil {
-		return "", err
-	}
-	return constraintPathMerge(target, p), nil
-}
-
 // constraintPathMerge is a shared function for creating constraint paths to
 // ensure uniformity, it is not meant to be called directly.
 func constraintPathMerge(target, subpath string) string {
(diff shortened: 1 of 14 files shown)


constraint/pkg/client/drivers/local/local.go
```diff
@@ -574,9 +583,50 @@ func (d *Driver) RemoveTemplate(ctx context.Context, templ *templates.Constraint
 	kind := templ.Spec.CRD.Spec.Names.Kind
 	namePrefix := createTemplatePath(targetHandler, kind)
 	_, err := d.deleteModules(namePrefix)
+	delete(d.templateHandler, templ.Spec.CRD.Spec.Names.Kind)
```
willbeason, line 586: This `delete` needs to be protected by a mutex since it is possible multiple callers could call `RemoveTemplate` simultaneously. It should be fine to use `d.modulesMux`

willbeason: Ditto other reads/writes for d.TemplateHandler.

Alternatively, you could use a sync.Map if you don't want to worry about this. While less performant than RWMutex, it's easier to read.

willbeason: Also feel free to just make a RWMutex just for d.templateHandler - any of these options should work.

constraint/pkg/client/client.go
```diff
@@ -492,27 +482,14 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
...
 		resp.Handled[target] = true
 	}
-
-	if len(errMap) == 0 {
-		c.constraints[constraint.GroupVersionKind().GroupKind()][subPath] = constraint.DeepCopy()
-		return resp, nil
-	}
-
-	return resp, &errMap
+	c.constraints[constraint.GroupVersionKind().GroupKind()][subPath] = constraint.DeepCopy()
```
willbeason, line 491: Make sure to guard this write with `c.mtx` so concurrent calls don't fight. Ditto other calls to c.constraints.

willbeason: Also - I'm pretty sure `c.constraints` isn't needed for anything any more. Can you add a TODO to the field declaration, and add an issue to remove the `constraints` field from `Client`?

becky-hd: Add TODO with link to #187, it would break `client.GetConstraint` if it has a usage.
The updates for c.constraints is guarded by c.mtx at the beginning of AddConstraint

constraint/pkg/client/drivers/local/local.go
maxsmythe: What happens if a template that has pre-existing constraints is deleted?

willbeason: Right now it is Client's responsibility to coordinate cascading deletes. So this doesn't change that flow. As a future PR, cascading deletes should be handled by Drivers.

constraint/pkg/client/drivers/local/local.go
```diff
@@ -664,6 +714,61 @@ func validateTargets(templ *templates.ConstraintTemplate) error {
...
+
+// createConstraintSubPath returns the key where we will store the constraint
+// for each target: cluster.<group>.<kind>.<name>.
+func createConstraintSubPath(constraint *unstructured.Unstructured) (string, error) {
+	if constraint.GetName() == "" {
+		return "", fmt.Errorf("%w: missing name", ErrInvalidConstraint)
+	}
+
+	gvk := constraint.GroupVersionKind()
+	if gvk.Group == "" {
```
willbeason, line 741: We should check that `gvk.Group == constraints.Group`, since that is the only valid value.

willbeason: (Also add a test for when gvk.Group is set, but to an incorrect value)

willbeason (approved):

maxsmythe (approved):
LGTM

willbeason (approved):

sozercan (approved):
LGTM

Issue #189, opened 2022-01-27
Make driver/handler cache operations atomic across Client


willbeason:
Not part of the current performance optimization - we currently do not add/remove objects atomically across drivers/targets.

Right now it is possible for us to end up in inconsistent states - where some Drivers/Handlers in a Client are able to cache objects but others are not. We currently opt to leave things in an inconsistent state rather than try to get back to a baseline.

This work is fairly complex to do generally - possibly needing its own design as there's a lot of edge cases to making these operations atomic.

willbeason:
Resolved for now since Templates can't change targets any more if they have any existing Constraints.

PR #190, opened 2022-01-27
Implement Target cache hook

willbeason:
Targets sometimes need to maintain state about the system. For example,
the Gatekeeper target needs to track the set of current Namespaces on
the cluster in order to properly match objects to Constraints when Audit
is called.

This commit adds a Cache interface which Targets may choose to
implement. If they implement this interface, Client attempts to add and
remove objects from the Target cache just as it does for Driver caches.

These operations are not atomic, so it is possible for systems to get
into an inconsistent state. There isn't a good solution to this now -
I've opened #189 to solve this in the future. The implications are quite
complex and there's a lot of edge cases.

This commit also modifies the test target handler matchers - they now
access the test target's cache in order to function. These matchers
aren't called yet - we don't want to break Gatekeeper since Gatekeeper
Golang matchers are not yet implemented.


willbeason:
Fixes #194

maxsmythe (approved):
LGTM

maxsmythe (approved):
LGTM

Issue #176, opened 2021-12-16
[client.Client] Record matchers for constraints


willbeason:
Prerequisites: #175 and #170 

Now that we've copied over matching logic into frameworks, and callers explicitly call "AddConstraint" and "RemoveConstraint", we can start doing bookkeeping in `local.Driver` which we need to have in order to move matching from Rego to Go. For this issue, we're just keeping a copy of the matchers for each Constraint.

Add a field to `local.Driver` called `matchers`. You can probably make this a map similar to the following:

```
type matcherKey struct {
  Kind, Name string
}
```

```
type Driver struct {
  ...
  matchers map[matcherKey]*Matcher
}
```

All that needs to happen is when Constraints are added to the Driver, we unmarshal the Matcher field into the Matcher type we copied over previously. Use the `bytes := json.Marshal(unstructured)` and `json.Unmarshal(bytes, matcher)` pattern used elsewhere to conver to the `Match` type.

We can't actually use these matchers yet since we haven't sharded the compilation environment - but we need these ready for when we do.

willbeason:
Decided to do this for client.Client since driver can be target-agnostic.

PR #193, opened 2022-02-01
Add type for matching Constraints

willbeason:
Prep work for compiler sharding. This adds a store of the matchers
corresponding to Constraints by-target and by-kind.

We could do constraint matching per-target, but it's better for us to
fail fast rather than partially review objects when we're just going to
return errors. Thus, ConstraintsFor iterates over all targets itself
rather than requiring its caller to.

The triple-nested maps are ugly - I would not need much convincing
to replace them with types.


willbeason:
My reason for the current implementation as maps-of-maps-of-maps is that this messiness is contained to the package - the type is fully contained in the `client` package and it is unexported. If we were to move this functionality to its own package, or to need this behavior to be extensible somehow, we'd want to use special types for the per-target and per-kind layers.

willbeason:
Fixes #176

constraint/pkg/client/matchers.go
```diff
@@ -0,0 +1,158 @@
...
+	if len(targetMatchers) == 0 {
+		delete(c.matchers, key.target)
+	} else {
+		c.matchers[key.target] = targetMatchers
+	}
+}
+
+// RemoveAll removes all Matchers for Constraints with kind.
+// Returns normally if no entry for the kind existed.
+func (c *constraintMatchers) RemoveAll(kind string) {
```
maxsmythe, line 94: nit: `RemoveKind`? To disambiguate vs. purging all matchers?

constraint/pkg/client/matchers.go
```diff
@@ -0,0 +1,158 @@
...
+// criteria.
+func (c *constraintMatchers) ConstraintsFor(review interface{}) (map[string]map[string][]string, error) {
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
+
+	result := make(map[string]map[string][]string)
+
+	errs := errors.ErrorMap{}
+
+	for target, targetMatchers := range c.matchers {
```
maxsmythe, line 130: What happens if a target is broadcasting that it doesn't deal with a given review via `HandleReview`?

https://github.com/open-policy-agent/frameworks/blob/4ee2501883a584d60b5756dc69888c547e867ba1/constraint/pkg/handler/handler.go#L75-L83

Also, `HandleReview()` should be called to preprocess the review to ensure it's in an appropriate format for `Match` (ex. AugmentedReview in Gatekeeper adding the namespace).

maxsmythe (commented):
One nit, one concern around where/when the `HandleReview` hook is called.

constraint/pkg/client/matchers.go
willbeason: Done.

constraint/pkg/client/matchers.go
willbeason: Done. Client now calls the matching logic per-target.

constraint/pkg/client/matchers.go
```diff
@@ -0,0 +1,173 @@
...
+		kindMatchers = make(map[string]constraints.Matcher)
+	}
+
+	kindMatchers[key.name] = matcher
+	t.matchers[key.kind] = kindMatchers
+}
+
+func (t *targetMatchers) Remove(key matcherKey) {
+	kindMatchers := t.matchers[key.kind]
+	if len(kindMatchers) == 0 {
```
maxsmythe, line 157: Should we do

```
kindMatchers, ok := t.matchers[key.kind]
```

and use `ok` here to check for existence?

That way we have GC regardless of whether the map happens to be empty?

constraint/pkg/client/matchers.go
```diff
@@ -0,0 +1,173 @@
...
+// Returns normally if no entry for the Constraint existed.
+func (c *constraintMatchers) Remove(key matcherKey) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	if len(c.matchers) == 0 {
+		return
+	}
+
+	target := c.matchers[key.target]
```
maxsmythe, line 58: Should we do

```
targe, ok := c.matchers[key.target]
```

and use `ok` here to check for existence?

That way we have GC regardless of whether the map happens to be empty?

constraint/pkg/client/matchers.go
willbeason: Done.

constraint/pkg/client/matchers.go
willbeason: Done.

constraint/pkg/client/matchers_test.go
```diff
@@ -0,0 +1,644 @@
...
+			name: "remove from empty target",
+			before: &constraintMatchers{
+				matchers: map[string]targetMatchers{
+					"foo": {},
+				},
+			},
+			key: matcherKey{target: "foo", kind: "bar", name: "qux"},
+			want: &constraintMatchers{
+				matchers: map[string]targetMatchers{
+					"foo": {},
```
maxsmythe, line 222: It looks like GC still isn't taking effect here?

maxsmythe (approved):
LGTM!

ritazh (approved):
lgtm

Issue #173, opened 2021-12-16
[local.Driver] Add methods AddCachedData and RemoveCachedData


willbeason:
Proposed signatures:
```
func (d *Driver) AddCachedData(ctx context.Context, obj interface{}) error
func (d *Driver) RemoveCachedData(ctx context.Context, obj interface{}) error
```

Under the hood these call PutData and DeleteData - AddInventory just allows Driver to choose where to store the data (note that the path string argument is missing).

This should have a corresponding change in Gatekeeper - you may find it helpful to [replace gatekeeper's dependency on frameworks with your local repo](https://thewebivore.com/using-replace-in-go-mod-to-point-to-your-local-module/) to figure this out.

For the purposes of this issue, it's fine to cast the `data` argument passed to `client.AddData` and return an error if it isn't the correct type. This will break `client.testHandler` - just use an Unstructured and pick fields that work similarly to how the `client.targetData` type works. Other than constructing an `Unstructured` instead of a `targetData`, this shouldn't require test code to change significantly.

willbeason:
Oh, and "Put Inventory" has to take interface{} and not Unstructured.

willbeason:
Per discussion, "AddCachedData" and "RemoveCachedData" are a better choice since it maps on to our general notion of "cached data" for referential constraints, whereas "inventory" is more of an implementation detail specific to a particular Driver's implementaiton.

Issue #174, opened 2021-12-16
[local.Driver] Make PutData and DeleteData private


willbeason:
Prerequisites: #173 and #170 

At this point, the only times PutData and DeleteData are called is in tests. This will flag tests you need to alter to use (and test) AddInventory/RemoveInventory and AddConstraint/RemoveConstraint appropriately.

PR #195, opened 2022-02-02
Implement Add/RemoveCachedData in drivers

davis-haba:
Implement Add/RemoveCachedData methods in driver. This moves path-building logic to drivers.

https://github.com/open-policy-agent/frameworks/issues/173


constraint/pkg/client/client.go
```diff
@@ -44,19 +44,12 @@ type Client struct {
...
-func createDataPath(target, subpath string) string {
-	subpaths := strings.Split(subpath, "/")
-	p := []string{"external", target}
-	p = append(p, subpaths...)
-
-	return "/" + path.Join(p...)
-}
-
-// AddData inserts the provided data into OPA for every target that can handle the data.
+// AddCachedData inserts the provided data into OPA for every target that can handle the data.
```
willbeason, line 47: Instead of "into OPA ..." - should be "every target and driver which caches data". For example, the gatekeeper cache doesn't (or rather, won't) use OPA and neither does the test target. Ditto for RemoveCachedData.

willbeason (approved):
LGTM except for comments

constraint/pkg/client/client.go
```diff
@@ -44,19 +44,12 @@ type Client struct {
...
-	return "/" + path.Join(p...)
-}
-
-// AddData inserts the provided data into OPA for every target that can handle the data.
+// AddCachedData inserts the provided data into every target and driver which caches data.
 // On error, the responses return value will still be populated so that
 // partial results can be analyzed.
-func (c *Client) AddData(ctx context.Context, data interface{}) (*types.Responses, error) {
+func (c *Client) AddCachedData(ctx context.Context, data interface{}) (*types.Responses, error) {
+	// TODO davis-haba delete comments when done
```
maxsmythe, line 51: TODO to delete comment?

constraint/pkg/client/drivers/interface.go
```diff
@@ -29,6 +29,8 @@ type Driver interface {
 	RemoveTemplate(ctx context.Context, ct *templates.ConstraintTemplate) error
 	PutData(ctx context.Context, path string, data interface{}) error
 	DeleteData(ctx context.Context, path string) (bool, error)
+	AddCachedData(ctx context.Context, target string, relPath string, data interface{}) error
```
maxsmythe, line 32: @willbeason 

Are PutData/DeleteData going to be removed once we no longer store constraints directly in OPA?

maxsmythe (approved):

constraint/pkg/client/drivers/interface.go
willbeason: Yes. See https://github.com/open-policy-agent/frameworks/issues/174

willbeason:
FYI the reason this breaks gatekeeper is the change to the client interface - this is expected.

constraint/pkg/client/client.go
```diff
@@ -91,9 +82,7 @@ func (c *Client) AddData(ctx context.Context, data interface{}) (*types.Response
 
 		// paths passed to driver must be specific to the target to prevent key
 		// collisions.
-		driverPath := createDataPath(target, relPath)
-		err = c.driver.PutData(ctx, driverPath, processedData)
-		if err != nil {
+		if err != c.driver.AddCachedData(ctx, target, relPath, processedData) {
```
ritazh, line 85: if err != nil ?

willbeason:
Absorbed these changes into mega PR which is now submitted.

willbeason:
Specifically: I copied these changes into #202

PR #196, opened 2022-02-04
Fix typos and minor mistakes

willbeason:
I ran inspect code to find typos and various unintentional bits of code.
All of these changes are minor and don't result in any behavioral
changes. Generally they show up as I'm editing code, and this makes them
go away.


sozercan (approved):
Thank you! LGTM

constraint/pkg/client/drivers/local/local.go
```diff
@@ -586,10 +586,13 @@ func (d *Driver) AddTemplate(templ *templates.ConstraintTemplate) error {
 
 // RemoveTemplate implements driver.Driver.
 // TODO: Cascading deletes.
-func (d *Driver) RemoveTemplate(ctx context.Context, templ *templates.ConstraintTemplate) error {
-	if err := validateTargets(templ); err != nil {
+func (d *Driver) RemoveTemplate(templ *templates.ConstraintTemplate) error {
+	if len(templ.Spec.Targets) != 1 {
```
maxsmythe, line 590: we probably want to return an error here to avoid failing silently if we forget to update this code when multi-target templates become a thing.

maxsmythe (commented):
once the error is added, LGTM

constraint/pkg/client/drivers/local/local.go
willbeason: If we return an error here, it's possible to get into a bad state where a template can't be deleted.

1. Add valid Template
2. Update Template to have multiple targets and wait for reconcile error
3. Delete Template

As-is, this code is probably broken if someone does a similar change. I'd rather investigate and resolve in a dedicated PR than add the test/logic necessary to deal with this case here. Thus, I'd rather not modify the code and possibly introduce a bug (or modify the behavior of an existing one) and feel it safer to leave as-is.

willbeason: Also - this will be made moot by compiler sharding since we'll fix this anyway.

willbeason: Filed #197 in the compiler sharding milestone.

willbeason: Also added TODO.

maxsmythe (approved):
LGTM

PR #198, opened 2022-02-08
Refactor out backend code

willbeason:
Note to reviewers: This changes how Client is instantiated, so it breaks
Gatekeeper's usage of it. This can easily be remedied with a small change
to the Gatekeeper repo.

Having the backend intermediate type for instantiating and using Client
adds complexity, code length, and unnecessary error paths. This change
makes using Client easier and streamlines options to a single type, all
specified when instantiating the Client.

This commit fully removes backend without breaking any existing
funcitonality.


maxsmythe (approved):
LGTM

sozercan (approved):
Thank you! LGTM

PR #200, opened 2022-02-16
Upgrade dependencies

willbeason:
The big one here is upgrading opa to v0.37.2 to fix the security vulnerability.


constraint/go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/open-policy-agent/frameworks/constraint
 
 go 1.16
 
+// Downgrade logr until we upgrade apimachinery.
+replace github.com/go-logr/logr => github.com/go-logr/logr v0.4.0
```
sozercan, line 6: what's the issue with apimachinery?

sozercan (approved):

willbeason:
I was able to upgrade to k8s 0.23.3 libraries, which gets us past the logr upgrade.

Open Telemetry has a habit of its dependencies getting out of sync, so I've added replace directives to keep them in lockstep. The go mod command isn't able to resolve these on its own.

As far as I can tell, we need to skip k8s 0.22.x since the Open Telemetry libraries are inconsistent across our dependencies.

willbeason:
The gatekeeper failure is expected - it needs to pin the otel dependencies in order for Go to resolve dependencies correctly.

sozercan (approved):
LGTM

maxsmythe (approved):
LGTM

constraint/go.mod
```diff
@@ -1,32 +1,104 @@
...
 	github.com/google/go-cmp v0.5.6
+	github.com/open-policy-agent/opa v0.37.2
+	github.com/spf13/cobra v1.3.0
+	github.com/spf13/pflag v1.0.5
+	golang.org/x/net v0.0.0-20211209124913-491a49abca63
+	k8s.io/apiextensions-apiserver v0.23.3
+	k8s.io/apimachinery v0.23.3
+	k8s.io/client-go v0.23.3
+	k8s.io/utils v0.0.0-20211116205334-6203023598ed
+	sigs.k8s.io/controller-runtime v0.11.1
```
ritazh, line 27: how does this impact the controller-runtime we use in gatekeeper? and will it work with the dynamiccache fork of controller-runtime https://github.com/open-policy-agent/gatekeeper/tree/master/third_party/sigs.k8s.io/controller-runtime?

willbeason: I'll test this locally to see if this upgrade breaks Gatekeeper.

willbeason: I've tested locally, and the upgrade works without impacting Gatekeeper. The dynamiccache fork works correctly.

ritazh (approved):
Thanks for the PR @willbeason ! 
LGTM

constraint/go.mod
sozercan: Would we need to upgrade our fork? @shomron

willbeason: Nope - we don't need to. I tested the upgrade and we don't have to modify our fork.

Issue #187, opened 2022-01-25
[client] clean constraints field in client


becky-hd:
Reevaluate if client.GetConstraint() is still needed, or the `constraint` field in client can be removed.

willbeason:
Looks like we'll still need Constraints in Client. In fact, it will become the source of truth for Constraints.

PR #202, opened 2022-03-02
perf: Implement compiler sharding

willbeason:
This is a massive PR - you may need more than one sitting to consume it.

You may find it helpful to skim https://docs.google.com/document/d/1ibCxaI-7HyWyDjQNL4iDRMHnrauufIMJ1D6LwIKdlsI/edit?usp=sharing again - the high-level design presented in the doc is largely unchanged.

Notable design changes:
- Queries are run per-Review per-Template. Before only a single query was run to review an object or run audit from cache.
- Audit from cache is removed. It isn't efficient to unmarshal everything from `data.inventory` to run Go matching logic, and in-rego matching logic costs too much time to run per-Template queries. Consumers of frameworks (e.g. Gatekeeper) will need to implement their own audit-from-cache. There are options for maintaining audit-from-cache in frameworks while keeping these performance gains, but they aren't pretty.
- Most built-in Rego code is removed - much of this orchestration has been moved to Go for performance reasons. The last bit is a requirement to keep users from needing to add it as boilerplate to their ConstraintTemplates.

Refactoring changes (non-behavioral)
- AllowedDataFields is moved to Driver as Client doesn't do anything with this data.
- Operations on Client are handled per-Template, per-Constraint, per-Target. In the future we could leverage this to allow for more highly-parallel mutations of the set of Constraints and Templates at once, but for now we're able to make updates to Client so quickly that this isn't an issue (discussion incoming with benchmarks). Mechanically, this is why we now have "client.go", "template_client.go", and "constraint_client.go" - each has concerns applicable to that level, and this allows separating out logic and reduces mixing layers of abstraction.

I've refrained from allowing much more parallelization in AddConstraint/AddTemplate operations (as in - adding Constraints and Templates simultaneously, or adding Constraints for different Templates simultaneously). While AddConstraint can be made more parallel, it comes at the cost of a lot of complexity and the absolute gains are not meaningful for known use cases (~5,000 Constraints added per second vs. ~20,000 Constraints added per second).

Update: Benchmark data is available here: https://docs.google.com/spreadsheets/d/1WYLMDdr9w9QwF9NGYvF1bzrYrKwF52Efj4bl7jaGMqA/edit?usp=sharing

### AddTemplate

Ignore multithreaded stats (crossed out). We aren't going to do this in this PR as it would hold this up. In the future, we can get an additional ~5x speedup in template compilation on top of what is shown here (e.g. 20x to 100x faster).

Compiling individual ConstraintTemplates is 2x-3x faster than before, from 4-5ms to 1.5-2ms.

Compiling 200 ConstraintTemplates is 10x-20x faster than before, from 3.4s-11s to 0.3s-0.5s.

There is improvement for all tested cases. Tests were run with both simple and complex rego code in Templates, so it is reasonable to assume all cases are improved.

### AddConstraint

AddConstraint is ~2x faster than before, so it is theoretically possible to add ~30,000 Constraints to a Gatekeeper installation per second. This is unlikely to be a bottleneck in any use case, so further improvements aren't necessary.

### Review

Review is complex.

**TLDR**: For most use cases, there are significant query improvements. The CPU cost of filtering out objects (and so not running Constraints) is dropped by 20x or more, so users should expect improvements in nearly all cases.

Note that for the data in the review benchmarks, I ran queries where every Constraint returned the same result. So "10 Templates & 100 Constraints / Success" means that all 100 Constraints are run against every object, and every Constraint resulted in no violation. These individual cases do not match reality where most objects are not matched by most Constraints. However, by testing these extreme cases individually, you can construct the expected improvement for various distributions of successes/failures/filters/autorjections.

I've taken data when running queries serially (in 1 thread) and in parallel (in 12 threads). The data for runtime in each page on the linked benchmark data is of _throughput_, not _spent CPU time_. In single-threaded mode these are equivalent, but this is not the case for multithreaded. For example, autorejecting a query for 1 Template/10 Constraints in one thread takes 18us, but only 4.3us with 12 threads. Effectively, this means that running in 12 threads means queries can be rejected at a rate of (1/4.3us) = 233,000/second, whereas running in a single thread the throughput is (1/18us) = 55,000/second.

Note that the throughput improvements are _universally greater_ than the CPU improvements. This is largely due to less read locking within Driver - all state is stored in the Rego storage (which has its own locking), so no synchronization is required directly by Driver.

For Constraints which are filtered out against a Review, or which autoreject a Review, speed is universally increased 10x-100x. Since this logic entirely happens in Go instead of Rego, we see a huge performance increase.

Otherwise, there is generally a 2x-3x speedup in Reviews where all Constraints are run for every object. The exception is the case where:
- there are many Templates
- there is exactly one Constraint per Template
- every Constraint matches every object on the cluster

This case is pathological - there shouldn't need to be so many Templates with a single Constraint, each matching every object in a Cluster. At these scales query time is still very reasonable - for example 18ms for 100 Constraints/100 Templates. We wouldn't expect there to be noticeable problems/degradation until a user had ~1,000 Constraints/1,000 Templates, when queries start to take longer than 200ms.


willbeason:
Also - breaking Gatekeeper is extremely expected so that failure can be ignored.

constraint/pkg/client/constraint_client.go
```diff
@@ -0,0 +1,76 @@
...
+	matcher, found := c.matchers[target]
+	if !found {
+		return nil
+	}
+
+	matches, err := matcher.Match(review)
+
+	switch {
+	case err != nil:
+		// Run the Constraint's enforcementAction since we were unable to determine
```
maxsmythe, line 41: nit: CF doesn't run enforcement actions, enforcement points would. Though we do return the enforcement action.

constraint/pkg/client/drivers/interface.go
```diff
@@ -5,35 +5,43 @@ import (
...
 	"github.com/open-policy-agent/frameworks/constraint/pkg/types"
+	"github.com/open-policy-agent/opa/storage"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-type QueryCfg struct {
-	TracingEnabled bool
+// A Driver implements Rego query execution of Templates and Constraints.
+type Driver interface {
+	AddTemplate(ct *templates.ConstraintTemplate) error
```
maxsmythe, line 14: nit: should we document what these interface functions do?

maxsmythe (commented):
Looking good! Initial round of feedback... got to drivers before I ran out of time (also have not looked at tests)

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,183 @@
...
-	template *templates.ConstraintTemplate
-	CRD      *apiextensions.CustomResourceDefinition
-	Targets  []string
-}
-
+// Client tracks ConstraintTemplates and Constraints for a set of Targets.
+// Allows validating reviews against Constraints.
+//
+// Threadsafe.
+// Assumes concurrent calls for the same object do not happen. For example -
```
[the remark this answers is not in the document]
willbeason: This note is completely unrelated to parallel compilation of Templates. Removing parallel computation of Templates does not remove this requirement.

This is a requirement as it is impossible for Frameworks to handle safely if this assumption is violated. There is _absolutely nothing_* that Frameworks can do to ensure proper behavior in the case that a caller sends a conflicting request to change an object before the previous operation has succeeded. Frameworks can't detect this scenario:

1. Start thread for `RemoveTemplate`
2. Thread for `RemoveTemplate` is frozen before any locking occurs (this can be the case even if the mutex lock is on the first line - e.g. while the function's stack is being instantiated)
3. Start thread for `AddTemplate`
4. `Add Template` gets a read lock.
5. Thread for `RemoveTemplate` wakes up, but can't do anything because it cannot acquire a lock.
6. `AddTemplate` completes successfully and releases the lock.
7. `RemoveTemplate` runs and deletes the Template.

The only safe way to handle this problem is to tell callers that they must make this guarantee.

*Under the current design. An alternative design would have Client run in its own thread, accepting work items synchronously, acquiring resource-specific locks, and launching work in goroutines, blocking callers until the work item is processed. This would be a vast departure from the design we currently have.

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,183 @@
...
+// concurrent calls to both update and remove a Template. Concurrent calls for
+// a Template and its Constraints are fine.
+//
+// Note that adding per-identifier locking would not fix this completely - the
+// thread for the first-sent call could be put to sleep while the second is
+// allowed to continue running. Thus, this problem can only safely be handled
```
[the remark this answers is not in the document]
willbeason: Added a full example where resource-specific locks results in incorrect behavior above.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -1,3 +1,42 @@
...
+const (
+	// EnforcementActionDeny indicates that if a review fails validation for a
+	// Constraint, that it should be rejected. Errors encountered running
+	// validation are treated as failing validation.
+	//
+	// This is the default EnforcementAction.
+	EnforcementActionDeny = "deny"
+	// EnforcementActionWarn indicates that a warning should be logged or printed
+	// if a review fails validation, or if the Constraint fails to run.
+	EnforcementActionWarn = "warn"
```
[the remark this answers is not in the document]
willbeason: Removed.

constraint/pkg/client/constraint_client.go
willbeason: Updated comment.

constraint/pkg/client/client.go
```diff
@@ -359,418 +187,404 @@ func (c *Client) AddTemplate(templ *templates.ConstraintTemplate) (*types.Respon
...
+	changed, err := template.AddConstraint(constraint)
+	if err != nil {
 		return resp, err
 	}
-	if err := c.driver.AddConstraint(ctx, constraint); err != nil {
-		return resp, err
+
+	if changed {
+		err = c.driver.AddConstraint(ctx, constraint)
+		if err != nil {
```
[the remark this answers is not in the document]
willbeason: Either:

1. We need to restart Policy Controller because the Rego inmem storage has gotten into a state where it no longer allows successful writes.
2. The context was cancelled (e.g. timed out)

constraint/pkg/client/client.go
```diff
@@ -359,418 +187,404 @@ func (c *Client) AddTemplate(templ *templates.ConstraintTemplate) (*types.Respon
...
+	err = c.driver.RemoveConstraint(ctx, constraint)
 	if err != nil {
-		return resp, err
+		return nil, err
 	}
-	if err := c.driver.RemoveConstraint(ctx, constraint); err != nil {
+
+	kind := constraint.GetKind()
+
+	c.mtx.Lock()
```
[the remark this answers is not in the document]
willbeason: Trying to evaluate against a Constraint no longer recognized by Driver evaluates to "no violation", which has the same end-result as deleting the Constraint (and so simply not running it).

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Also - it's okay for the state to be kept consistent (Constraint in both Driver and per-Template cache) if removal from the driver fails due to context cancellation.

willbeason:
I'm taking benchmarks now. See data here: https://docs.google.com/spreadsheets/d/1WYLMDdr9w9QwF9NGYvF1bzrYrKwF52Efj4bl7jaGMqA/edit?usp=sharing

I'll post a summary once I've taken all of the benchmarks for AddTemplate/AddConstraint/Review

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: That makes sense. My main concern would be that concurrent requests cannot put the framework into an inconsistent state (e.g., parts of its state think the template has been added, parts of its state think it's been removed).

Clients should be responsible for avoiding race conditions in their own logic (which is what you describe), we should be responsible for maintaining consistent state (or raising a panic when we've failed to do so), regardless of how we're called.

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,183 @@
...
+		}
+		sort.Strings(oldTargets)
+		sort.Strings(newTargets)
 
-// createRawTemplateArtifacts creates the "free" artifacts for a template, avoiding more
-// complex tasks like rewriting Rego. Provides minimal validation.
-func (c *Client) createRawTemplateArtifacts(templ *templates.ConstraintTemplate) (*rawCTArtifacts, error) {
-	if templ.GetName() == "" {
-		return nil, fmt.Errorf("%w: missing name", clienterrors.ErrInvalidConstraintTemplate)
+		for i, target := range oldTargets {
```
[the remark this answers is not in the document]
maxsmythe: +1 length check makes this a strictly-no-change requirement, which makes sense.

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,183 @@
...
-		template: cpy,
-		CRD:      basicArtifacts.crd,
-		Targets:  []string{basicArtifacts.targetHandler.GetName()},
-	}
+func getTargetName(templ *templates.ConstraintTemplate) (string, error) {
+	targets := templ.Spec.Targets
 
-	if _, ok := c.constraints[basicArtifacts.gk]; !ok {
-		c.constraints[basicArtifacts.gk] = make(map[string]*unstructured.Unstructured)
+	switch len(targets) {
```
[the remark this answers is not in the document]
maxsmythe: But the error texts are exactly the same?

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: IIRC at one point we talked about panic-ing if modifying the driver's cache fails? Mainly b/c such failures are unexpected (need to restart PC) and in the case of multi-target it makes it hard/impossible to have consistent state.

constraint/pkg/client/client.go
```diff
@@ -359,418 +187,404 @@ func (c *Client) AddTemplate(templ *templates.ConstraintTemplate) (*types.Respon
...
 	resp := types.NewResponses()
-	entry, err := c.getTemplateEntry(constraint, false)
+
+	err := validateConstraintMetadata(constraint)
 	if err != nil {
 		return resp, err
 	}
-	subPath, err := createConstraintSubPath(constraint)
+
+	err = c.driver.RemoveConstraint(ctx, constraint)
```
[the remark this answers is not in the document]
maxsmythe: Per above, should we build in the expectation that drivers can always cache/uncache data and should panic otherwise?

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: Will that always be true? Absent significant performance gains, consistency seems safer here.

willbeason:
Benchmarks are in the above spreadsheet now, but I need to write about how to interpret them. Those will be coming tomorrow.

constraint/pkg/client/drivers/local/args.go
```diff
@@ -1,114 +1,136 @@
...
 		// adding external_data builtin otherwise capabilities get overridden
 		// if a capability, like http.send, is disabled
 		if d.providerCache != nil {
-			d.capabilities.Builtins = append(d.capabilities.Builtins, &ast.Builtin{
+			newBuiltin := &ast.Builtin{
 				Name: "external_data",
 				Decl: opatypes.NewFunction(opatypes.Args(opatypes.A), opatypes.A),
-			})
+			}
+			d.compilers.capabilities.Builtins = append(d.compilers.capabilities.Builtins, newBuiltin)
```
maxsmythe, line 40: It looks like `d.compilers.capabilities` could be nil unless the `DisableBuiltins` option is set?

constraint/pkg/client/drivers/local/compilers.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
+	"github.com/open-policy-agent/frameworks/constraint/pkg/regorewriter"
+	"github.com/open-policy-agent/opa/ast"
+)
+
+// Compilers is a threadsafe store of Compilers for ConstraintTemplates.
+type Compilers struct {
+	mtx sync.RWMutex
+
+	// compilers is a map from target name to a map from kind to compiler.
```
maxsmythe, line 17: nit: kind -> constraint template kind

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -0,0 +1,435 @@
...
+	// once per call to Query instead of once per compiler.
+	reviewMap, err := toInterfaceMap(review)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for kind, kindConstraints := range constraintsByKind {
+		compiler := d.compilers.getCompiler(target, kind)
+		if compiler == nil {
+			// The Template was just removed, so ignore these Constraints.
```
maxsmythe, line 244: Is this a state it's possible to be in? client acquires a read lock before it gathers the list of constraints.

If not, we may want to raise an error to make the unexpected state more visible

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -0,0 +1,435 @@
...
+
+	return string(b), nil
+}
+
+// templateLibPrefix returns the new lib prefix for the libs that are specified in the CT.
+func templateLibPrefix(name string) string {
+	return fmt.Sprintf("libs.%s", name)
+}
+
+// createTemplatePath returns the package path for a given template: templates.<target>.<name>.
```
maxsmythe, line 327: nit: comment appears to be stale (path has changed)

maxsmythe (commented):
Took a pass through everything but tests

constraint/pkg/client/drivers/local/driver.go
willbeason: Fixed.

constraint/pkg/client/drivers/local/driver.go
willbeason: Modified to not allow concurrent changes to Client.

willbeason: Done.

constraint/pkg/client/drivers/local/compilers.go
willbeason: Changed to "Constraint kind" as all ConstraintTemplates have the same "kind".

constraint/pkg/client/drivers/local/args.go
willbeason: `d.compilers.capabilities` is defaulted in Defaults() if unset:

```
if d.compilers.capabilities == nil {
	d.compilers.capabilities = ast.CapabilitiesForThisVersion()
}
```

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Changed Client to not allow concurrent updates or compilation. I don't want to hold up the PR on this - speeding up compilation by 20x is significant on it's own. While it would be nice to get the 5x additional speedup to 100x, we can do that later (and do it right).

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Made Client no longer allow concurrent updates. While we talked about panicking if modifying the driver's cache fails, it can fail for reasons either meaning Frameworks should be restarted or due to a simple context cancellation/timeout. These modifications aren't necessary for this PR and can be done later. This PR doesn't change how that situation is handled from the existing behavior of just reporting when this happens.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Closing this thread since the explanations/arguments/solutions here are identical to AddConstraint.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Made Client not allow concurrent updates, so this isn't an issue.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: I'd rather leave behavior as-is in regards to failing to update the OPA storage, rather than potentially introduce a case where Gatekeeper constantly restarts due to context cancellations being much more common than we anticipate for certain users.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Ah, I missed that. Done.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
willbeason: Made Client no longer allow concurrent updates.

constraint/pkg/client/template_client.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/open-policy-agent/frameworks/constraint/pkg/handler"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// templateClient handles per-ConstraintTemplate operations.
+//
+// Not threadsafe.
+type templateClient struct {
+	// targets are the Targets which
```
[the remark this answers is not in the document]
willbeason: Done.

constraint/pkg/client/drivers/local/rego.go
```diff
@@ -0,0 +1,49 @@
...
+package hooks
+
+# Determine if the object under review violates any passed Constraints.
+violation[response] {
+  # Iterate over all keys to Constraints in storage.
+  key := input.constraints[_]
+
+  # Construct the input object from the Constraint and temporary object in storage.
+  # Silently exits if the Constraint no longer exists.
+	inp := {
```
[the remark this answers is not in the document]
willbeason: Done.

constraint/pkg/handler/key.go
```diff
@@ -0,0 +1,9 @@
+package handler
+
+import "strings"
+
+type StoragePath []string
+
+func (p StoragePath) String() string {
+	return strings.Join(p, "/")
```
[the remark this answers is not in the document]
willbeason: Changed to use Rego Storage logic which escapes properly.

constraint/pkg/client/drivers/interface.go
willbeason: Done.

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
 
-			if cache != nil {
-				cache.Remove(relPath)
-			}
-			continue
-		}
+	c.mtx.RLock()
```
maxsmythe, line 57: Do we need the lock here?

Is all we're doing reading the target handler cache? That should be constant post start-up.

maxsmythe (approved):
LGTM, minor nits.

maxsmythe:
Locking question for creating CRD most significant nit.

constraint/pkg/client/client.go
willbeason: Good catch! Done.

maxsmythe (approved):
Still LGTM, with the caveat we should rename the per-target `data.inventory[<target name>]` cache (e.g. previously `data.external`), to minimize risk that future changes will make more cached data than we intended be visible for things like namespace-scoped constraints.

maxsmythe (approved):

constraint/pkg/apis/constraints/apis.go
```diff
@@ -1,3 +1,39 @@
 package constraints
 
+import (
+	"errors"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
 const Group = "constraints.gatekeeper.sh"
```
sozercan, line 10: nit: can we add this to const group?

sozercan (approved):
Thank you!! LGTM

sozercan:
@willbeason looks like unit tests were timing out, re-ran it. do we need to increase the timeout value or is it a problem with code/tests?

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,174 @@
...
+	// Does not require mutex locking as Driver is threadsafe.
+	driver drivers.Driver
+	// targets are the targets supported by this Client.
+	// Assumed to be constant after initialization.
 	targets map[string]handler.TargetHandler
 
-	// mtx guards access to both templates and constraints.
-	mtx       sync.RWMutex
-	templates map[templateKey]*templateEntry
-	// TODO: https://github.com/open-policy-agent/frameworks/issues/187
```
ritazh, line 42: From https://github.com/open-policy-agent/frameworks/issues/187
> Looks like we'll still need Constraints in Client. In fact, it will become the source of truth for Constraints.

why is constraints removed from Client?

constraint/pkg/client/client.go
```diff
@@ -1,355 +1,174 @@
...
-				// Use a different key than the driver to avoid clobbering errors.
-				errMap[target] = err
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
 
-				continue
-			}
-		}
+	// Return immediately if no change.
+	targetName, err := getTargetName(templ)
```
ritazh, line 75: check for ` != ""`

constraint/pkg/client/client.go
```diff
@@ -359,418 +178,402 @@ func (c *Client) AddTemplate(templ *templates.ConstraintTemplate) (*types.Respon
...
-		return resp, fmt.Errorf("creating Constraint subpath: %w", err)
+		return resp, err
 	}
 
-	// return immediately if no change
-	cached, err := c.getConstraintNoLock(constraint)
-	if err == nil && constraintlib.SemanticEqual(cached, constraint) {
-		for _, target := range entry.Targets {
-			resp.Handled[target] = true
+	if changed {
```
ritazh, line 259: check `!= nil` ?

constraint/pkg/apis/constraints/apis.go
willbeason: Done!

willbeason:
> @willbeason looks like unit tests were timing out, re-ran it (edit: timed out again). do we need to increase the timeout value or is it a problem with code/tests?
> 
> here's the full output
> 
> ```
> go test ./pkg/... -coverprofile cover.out
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/apis     [no test files]
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/apis/constraints [no test files]
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/apis/externaldata/v1alpha1       6.551s  coverage: 2.8% of statements
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates   [no test files]
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1        6.532s  coverage: 45.9% of statements
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1alpha1  6.438s  coverage: 45.9% of statements
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/apis/templates/v1beta1   6.498s  coverage: 45.9% of statements
> coverage: 33.2% of statements
> panic: test timed out after 10m0s
> 
> goroutine 107 [running]:
> testing.(*M).startAlarm.func1()
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1788 +0x8e
> created by time.goFunc
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/time/sleep.go:180 +0x31
> 
> goroutine 1 [chan receive, 10 minutes]:
> testing.(*T).Run(0xc0005824e0, {0x14ada09, 0x46f053}, 0x152bf18)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1307 +0x375
> testing.runTests.func1(0xc0005824e0)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1598 +0x6e
> testing.tRunner(0xc0005824e0, 0xc00073fce0)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1259 +0x102
> testing.runTests(0xc000050200, {0x219f220, 0x18, 0x18}, {0x48daed, 0x14a2da8, 0x21b7ec0})
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1596 +0x43f
> testing.(*M).Run(0xc000050200)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1504 +0x51d
> main.main()
>         _testmain.go:161 +0x1f5
> 
> goroutine 19 [chan receive]:
> k8s.io/klog/v2.(*loggingT).flushDaemon(0x0)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/k8s.io/klog/v2/klog.go:1283 +0x6a
> created by k8s.io/klog/v2.init.0
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/k8s.io/klog/v2/klog.go:420 +0xfb
> 
> goroutine 12 [chan receive]:
> github.com/golang/glog.(*loggingT).flushDaemon(0xc00018d740)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/github.com/golang/glog/glog.go:882 +0x6a
> created by github.com/golang/glog.init.0
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/github.com/golang/glog/glog.go:410 +0x1c5
> 
> goroutine 103 [chan receive, 10 minutes]:
> testing.(*T).Run(0xc000603380, {0x14a0a6c, 0x149ef92}, 0xc0002cac80)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1307 +0x375
> github.com/open-policy-agent/frameworks/constraint/pkg/client_test.TestClient_AddTemplate(0xc000603380)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/client_test.go:376 +0x189c
> testing.tRunner(0xc000603380, 0x152bf18)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1259 +0x102
> created by testing.(*T).Run
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1306 +0x35a
> 
> goroutine 104 [semacquire, 10 minutes]:
> sync.runtime_SemacquireMutex(0xc000296400, 0x60, 0xc0001c1660)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/runtime/sema.go:71 +0x25
> sync.(*RWMutex).RLock(...)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/sync/rwmutex.go:63
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.(*Driver).getStorage(0xc000225200, {0x16a5060, 0xc0001aa000}, {0x149ef92, 0xb})
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver.go:526 +0x97
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.(*Driver).removeData(0x12ddba0, {0x16a5060, 0xc0001aa000}, {0x149ef92, 0xc0001c1680}, {0xc000358360, 0x2, 0x2})
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver.go:494 +0x5b
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.(*Driver).RemoveTemplate(0xc000225200, {0x16a5060, 0xc0001aa000}, 0x0)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver.go:105 +0x1d7
> github.com/open-policy-agent/frameworks/constraint/pkg/client.(*Client).RemoveTemplate(0xc00013fb00, {0x16a5060, 0xc0001aa000}, 0xc000156180)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/client.go:184 +0x155
> github.com/open-policy-agent/frameworks/constraint/pkg/client_test.TestClient_AddTemplate.func1(0xc000603520)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/client_test.go:438 +0x737
> testing.tRunner(0xc000603520, 0xc0002cac80)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1259 +0x102
> created by testing.(*T).Run
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1306 +0x35a
> FAIL    github.com/open-policy-agent/frameworks/constraint/pkg/client   600.016s
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/client/clienttest        [no test files]
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/client/clienttest/cts    [no test files]
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/client/crds      0.020s  coverage: 89.6% of statements
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers   [no test files]
> coverage: 49.1% of statements
> panic: test timed out after 10m0s
> 
> goroutine 83 [running]:
> testing.(*M).startAlarm.func1()
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1788 +0x8e
> created by time.goFunc
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/time/sleep.go:180 +0x31
> 
> goroutine 1 [chan receive, 9 minutes]:
> testing.(*T).Run(0xc0000c4680, {0x130c5e3, 0x46f053}, 0x1380fb8)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1307 +0x375
> testing.runTests.func1(0xc0000c4680)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1598 +0x6e
> testing.tRunner(0xc0000c4680, 0xc00073fce0)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1259 +0x102
> testing.runTests(0xc000416080, {0x1e80000, 0x9, 0x9}, {0x48daed, 0x12fdb33, 0x1e99860})
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1596 +0x43f
> testing.(*M).Run(0xc000416080)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1504 +0x51d
> main.main()
>         _testmain.go:117 +0x1f5
> 
> goroutine 19 [chan receive]:
> k8s.io/klog/v2.(*loggingT).flushDaemon(0x0)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/k8s.io/klog/v2/klog.go:1283 +0x6a
> created by k8s.io/klog/v2.init.0
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/k8s.io/klog/v2/klog.go:420 +0xfb
> 
> goroutine 20 [chan receive]:
> github.com/golang/glog.(*loggingT).flushDaemon(0x0)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/github.com/golang/glog/glog.go:882 +0x6a
> created by github.com/golang/glog.init.0
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/vendor/github.com/golang/glog/glog.go:410 +0x1c5
> 
> goroutine 16 [chan receive, 9 minutes]:
> testing.(*T).Run(0xc0006031e0, {0x12fdc4b, 0xc0001404a0}, 0xc0002a5680)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1307 +0x375
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.TestDriver_RemoveTemplates(0x0)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver_unit_test.go:158 +0x1dd
> testing.tRunner(0xc0006031e0, 0x1380fb8)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1259 +0x102
> created by testing.(*T).Run
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1306 +0x35a
> 
> goroutine 82 [semacquire, 9 minutes]:
> sync.runtime_SemacquireMutex(0x12, 0x0, 0x8)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/runtime/sema.go:71 +0x25
> sync.(*RWMutex).RLock(...)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/sync/rwmutex.go:63
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.(*Driver).getStorage(0xc0001b6cf0, {0x14cd8e0, 0xc000140000}, {0x12f0a22, 0x3})
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver.go:526 +0xa5
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.(*Driver).removeData(0x1162260, {0x14cd8e0, 0xc000140000}, {0x12f0a22, 0xc0001b6cf0}, {0xc00052db20, 0x2, 0x2})
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver.go:494 +0x65
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.(*Driver).RemoveTemplate(0xc0001b6cf0, {0x14cd8e0, 0xc000140000}, 0x0)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver.go:105 +0x1ea
> github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local.TestDriver_RemoveTemplates.func1(0xc000603380)
>         /home/sozercan/go/src/github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local/driver_unit_test.go:176 +0x378
> testing.tRunner(0xc000603380, 0xc0002a5680)
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1259 +0x102
> created by testing.(*T).Run
>         /home/linuxbrew/.linuxbrew/Cellar/go/1.17.8/libexec/src/testing/testing.go:1306 +0x35a
> FAIL    github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local     600.019s
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/remote    0.010s  coverage: 19.3% of statements
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/client/errors    [no test files]
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/core/constraints 0.044s  coverage: 100.0% of statements
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/core/templates   [no test files]
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/externaldata     0.025s  coverage: 61.4% of statements
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/handler  [no test files]
> ?       github.com/open-policy-agent/frameworks/constraint/pkg/handler/handlertest      [no test files]
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/regorewriter     0.055s  coverage: 71.8% of statements
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/schema   0.024s  coverage: 50.0% of statements
> ok      github.com/open-policy-agent/frameworks/constraint/pkg/types    0.024s  coverage: 13.6% of statements
> FAIL
> Makefile:15: recipe for target 'native-test' failed
> make: *** [native-test] Error 1
> ```

It was a problem with the code - I didn't lock properly. It's fixed, and I've moved that part of the code to it's own class so future devs will be unlikely to make the same mistake I did. Otherwise the logic is unchanged - just mutexes moved around.

constraint/pkg/client/client.go
willbeason: Constraints aren't removed from Client. They're now managed by `templateClient`, a subfield of `Client`.

constraint/pkg/client/client.go
willbeason: What are you suggesting be checked for nil?

ritazh: argh I was looking at the AddConstraint from Client instead of the one from template. so confusing when they all have the same name!

constraint/pkg/client/client.go
```diff
@@ -359,418 +183,402 @@ func (c *Client) AddTemplate(templ *templates.ConstraintTemplate) (*types.Respon
...
 
-	resp := types.NewResponses()
-	entry, err := c.getTemplateEntry(constraint, false)
+	err := c.validateConstraint(constraint)
 	if err != nil {
 		return resp, err
 	}
 
-	subPath, err := createConstraintSubPath(constraint)
+	kind := constraint.GetKind()
```
ritazh, line 252: check for `kind != ""`

ritazh (approved):
LGTM

constraint/pkg/client/client.go
willbeason: We do this in `validateConstraintMetadata` just before, and this code path has a unit test.

PR #203, opened 2022-03-17
Make library prefix constant

willbeason:
Instead of using a variable template-specific location for libraries, use a fixed location so that users can refer to libraries by _either_ their fully qualified name _or_ their local name.

This behavior didn't have a complex enough test to trigger a bug found in Gatekeeper while merging the compiler sharding PR. This fixes the bug and adds a sufficiently-complex ConstraintTemplate which triggers the library import bug.


maxsmythe (approved):
LGTM

willbeason:
This bugfix also reduces compilation time by another 20%+ and reduced memory usage by 30%+

sozercan (approved):
Thanks! LGTM

PR #207, opened 2022-04-01
Upgrade opa to v0.39.0

willbeason:
Upgrade OPA version


willbeason:
FYI we can't remove the replace directives until k8s.io/component-base upgrades its opentelemetry version. For now we've gotten lucky as while OPA and k8s.io/component-base use incompatible versions, they don't do this in a way that results in compilation errors. We won't always be this fortunate.

srenatus (approved):

maxsmythe (approved):
LGTM

PR #208, opened 2022-04-04
Remove direct storage.Path requirement in handler

willbeason:
As-is requires a direct dependency of all callers on OPA. This PR
changes the TargetHandler interface to accept storage path as a slice of
strings instead of storage.Path.

Functionally this changes nothing.


maxsmythe (approved):
LGTM, thanks for this!

PR #216, opened 2022-04-06
perf: Don't cache Constraint status

### LINKED ISSUE #209: No Need to Cache Constraint Status
We should strip the constraint status from the constraint in AddConstraint() in order to lower the footprint when constraints have large status size (e.g. many audit results):

https://github.com/open-policy-agent/frameworks/blob/8066162cf5e2ba18308c339e2ab57f834948db22/constraint/pkg/client/client.go#L238-L276

willbeason:
Constraint status is ephemeral and callers should not be relying on our
cached version of the status field. Also, status can be very large, and
caching it wastes space.

Note that the Driver already drops `status` as it only preserves `spec.parameters`.

Fixes #209

### DIFF
--- a/constraint/pkg/client/drivers/local/driver.go
+++ b/constraint/pkg/client/drivers/local/driver.go
@@ -103,6 +103,7 @@ func (d *Driver) RemoveTemplate(ctx context.Context, templ *templates.Constraint
 // AddConstraint adds Constraint to Rego storage. Future calls to Query will
 // be evaluated against Constraint if the Constraint's key is passed.
 func (d *Driver) AddConstraint(ctx context.Context, constraint *unstructured.Unstructured) error {
+	// Note that this discards "status" as we're only copy spec.parameters.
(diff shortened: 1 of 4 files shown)


constraint/pkg/client/drivers/local/driver.go
```diff
@@ -103,6 +103,7 @@ func (d *Driver) RemoveTemplate(ctx context.Context, templ *templates.Constraint
 // AddConstraint adds Constraint to Rego storage. Future calls to Query will
 // be evaluated against Constraint if the Constraint's key is passed.
 func (d *Driver) AddConstraint(ctx context.Context, constraint *unstructured.Unstructured) error {
+	// Note that this discards "status" as we're only copy spec.parameters.
```
maxsmythe, line 106: nit: we only copy

maxsmythe (approved):
LGTM!

constraint/pkg/client/drivers/local/driver.go
willbeason: Done.

sozercan (approved):
LGTM

PR #226, opened 2022-05-18
feat: TLS support for External Data Providers

chewong:
Design doc: https://docs.google.com/document/d/1GjV3WeC2bgQ3j37_mMpY9hr7YOAqzSJ6jDSu-DVrcmU/edit#

This PR adds two new fields to the Provider CRD - caBundle and insecureTLSSkipVerify. If the provider is using HTTPS scheme for their provider and presents a valid caBundle, we will inject it into our HTTP client before making an external data request. We will also inject Gatekeeper's certificate into our HTTP client in case the provider wants to establish mutual trust.

### DIFF
--- a/constraint/pkg/client/drivers/local/args.go
+++ b/constraint/pkg/client/drivers/local/args.go
@@ -2,6 +2,7 @@ package local
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/open-policy-agent/frameworks/constraint/pkg/client/errors"
 	"github.com/open-policy-agent/frameworks/constraint/pkg/externaldata"
@@ -45,6 +46,10 @@ func Defaults() Arg {
 
 		if d.sendRequestToProvider == nil {
 			d.sendRequestToProvider = externaldata.DefaultSendRequestToProvider
+		}
+
+		if d.fs == nil {
+			d.fs = os.DirFS("/")
 		}
 
 		return nil
--- a/constraint/pkg/client/drivers/local/args.go
+++ b/constraint/pkg/client/drivers/local/args.go
@@ -129,6 +129,14 @@ func AddClientTLSKeyPair(certFile, keyFile string) Arg {
 	}
 }
 
+func EnableExternalDataClientAuth() Arg {
+	return func(d *Driver) error {
+		d.enableExternalDataClientAuth = true
+
+		return nil
+	}
+}
+
 // Externs sets the fields under `data` that Rego in ConstraintTemplates
 // can access. If unset, all fields can be accessed. Only fields recognized by
 // the system can be enabled.
--- a/constraint/pkg/client/drivers/local/args.go
+++ b/constraint/pkg/client/drivers/local/args.go
@@ -115,6 +115,15 @@ func DisableBuiltins(builtins ...string) Arg {
 	}
 }
 
+func AddClientTLSKeyPair(certFile, keyFile string) Arg {
+	return func(d *Driver) error {
+		d.clientCertFile = certFile
+		d.clientKeyFile = keyFile
+
+		return nil
+	}
+}
+
 // Externs sets the fields under `data` that Rego in ConstraintTemplates
 // can access. If unset, all fields can be accessed. Only fields recognized by
 // the system can be enabled.
--- a/constraint/pkg/client/drivers/local/args.go
+++ b/constraint/pkg/client/drivers/local/args.go
@@ -2,7 +2,6 @@ package local
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/open-policy-agent/frameworks/constraint/pkg/client/errors"
 	"github.com/open-policy-agent/frameworks/constraint/pkg/externaldata"
@@ -10,6 +9,7 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/topdown/print"
 	opatypes "github.com/open-policy-agent/opa/types"
+	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 )
 
 type Arg func(*Driver) error
@@ -46,10 +46,6 @@ func Defaults() Arg {
 
 		if d.sendRequestToProvider == nil {
 			d.sendRequestToProvider = externaldata.DefaultSendRequestToProvider
-		}
-
-		if d.fs == nil {
-			d.fs = os.DirFS("/")
 		}
 
 		return nil
@@ -120,10 +116,9 @@ func DisableBuiltins(builtins ...string) Arg {
 	}
 }
 
-func AddClientTLSKeyPair(certFile, keyFile string) Arg {
+func AddExternalDataClientCertWatcher(clientCertWatcher *certwatcher.CertWatcher) Arg {
 	return func(d *Driver) error {
-		d.clientCertFile = certFile
-		d.clientKeyFile = keyFile
+		d.clientCertWatcher = clientCertWatcher
 
 		return nil
 	}
--- a/constraint/pkg/client/drivers/local/builtin.go
+++ b/constraint/pkg/client/drivers/local/builtin.go
@@ -1,7 +1,6 @@
 package local
 
 import (
-	"crypto/tls"
 	"net/http"
 
 	"github.com/open-policy-agent/frameworks/constraint/pkg/externaldata"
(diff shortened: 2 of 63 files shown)


constraint/pkg/client/drivers/local/builtin.go
```diff
@@ -20,7 +21,12 @@ func externalDataBuiltin(d *Driver) func(bctx rego.BuiltinContext, regorequest *
 			return externaldata.HandleError(http.StatusBadRequest, err)
 		}
 
-		externaldataResponse, statusCode, err := d.sendRequestToProvider(bctx.Context, &provider, regoReq.Keys)
+		clientCert, err := tls.LoadX509KeyPair(d.clientCertFile, d.clientKeyFile)
```
maxsmythe, line 24: What happens if the contents of the cert file changes? do we need to watch the file like controller-runtime's webserver does?

constraint/pkg/externaldata/cache.go
```diff
@@ -69,6 +73,32 @@ func isValidURL(url string) bool {
 	return true
 }
 
+func isValidCABundle(provider *v1alpha1.Provider) error {
+	if provider.Spec.InsecureTLSSkipVerify {
+		if provider.Spec.CABundle != "" {
+			return fmt.Errorf("insecureTLSSkipVerify is set to true but caBundle is not empty")
```
maxsmythe, line 79: Is this necessarily bad? Might users be testing whether the CA bundle is bad in this scenario?

constraint/pkg/externaldata/cache.go
```diff
@@ -69,6 +73,32 @@ func isValidURL(url string) bool {
...
+	u, err := url.Parse(provider.Spec.URL)
+	if err != nil {
+		return err
+	}
+
+	switch u.Scheme {
+	case "http":
+		return fmt.Errorf("only HTTPS scheme is supported for Providers. To enable HTTP scheme, set insecureTLSSkipVerify to true")
+	case "https":
+		if provider.Spec.CABundle == "" {
```
maxsmythe, line 94: Should we attempt to parse/load the CA bundle? Some checks we do for CA certs can be found in cert controller.

maxsmythe (commented):
Couple comments, but mostly LGTM

constraint/pkg/client/drivers/local/builtin.go
chewong: Since we didn't cache the content, it loads the files every time `external_data` is called so I don't think we need to watch it. Any changes to the files should be picked up by the subsequent call.

constraint/pkg/externaldata/cache.go
chewong: Good point. I removed this condition to allow users to test their caBundle while InsecureTLSSkipVerify is enabled.

constraint/pkg/client/drivers/local/builtin.go
maxsmythe: Ah, I see. Are there concerns about the extra latency this may induce? Disk access can be slow, particularly if its a spinning disk.

chewong: Let me do some testing on this. I can also introduce a flag to disable this feature by default and user can enable it if they want mTLS between Gateekeeper and providers.

maxsmythe: +1 to being able to flag-gate this. That'd also be useful for people who are not interested in mTLS.

constraint/pkg/client/drivers/local/builtin.go
```diff
@@ -20,7 +21,12 @@ func externalDataBuiltin(d *Driver) func(bctx rego.BuiltinContext, regorequest *
...
+		if err != nil {
```
willbeason, line 25: Add tests for these various cases to ensure they work properly.

constraint/pkg/externaldata/cache.go
```diff
@@ -69,6 +76,53 @@ func isValidURL(url string) bool {
...
+			return fmt.Errorf("failed to base64 decode caBundle: %w", err)
+		}
+
+		caDer, _ := pem.Decode(caPem)
+		if caDer == nil {
+			return fmt.Errorf("bad caBundle")
+		}
+
+		_, err = x509.ParseCertificate(caDer.Bytes)
+		if err != nil {
```
willbeason, line 92: Add test that ensures we fail as expected when caDer.Bytes fails to parse.

Also - add comment on why we're doing this parsing. Since it isn't obvious _why_ we're doing this, it would be easy for a future code editor to remove this check. No tests would fail and there's no comment explaining why this check is necessary.

constraint/pkg/externaldata/cache.go
```diff
@@ -69,6 +76,53 @@ func isValidURL(url string) bool {
...
+			return fmt.Errorf("failed to parse caBundle: %w", err)
+		}
+
+		return nil
+	}
+
+	u, err := url.Parse(provider.Spec.URL)
```
willbeason, line 99: Add test for if `provider.Spec.URL` fails to parse

constraint/pkg/externaldata/cache.go
```diff
@@ -69,6 +76,53 @@ func isValidURL(url string) bool {
...
+	case "http":
+		if !provider.Spec.InsecureTLSSkipVerify {
+			return fmt.Errorf("only HTTPS scheme is supported for Providers. To enable HTTP scheme, set insecureTLSSkipVerify to true")
+		}
+		if provider.Spec.CABundle != "" {
+			if err := verify(provider.Spec.CABundle); err != nil {
+				return err
+			}
+		}
+	case "https":
```
willbeason, line 114: Use string constants for `"http"` and `"https"` to prevent future typos

constraint/pkg/client/drivers/local/builtin.go
willbeason, line 24: Add tests for this functionality. As-is would likely be annoying, so you'll likely want to use `fs.FS` so that you can test this without needing to use actual files.

(Such tests would require switching to `tls.X509KeyPair` instead of `tls.LoadX509KeyPair`)

willbeason (changes requested):

constraint/pkg/externaldata/cache.go
chewong: > Add test that ensures we fail as expected when caDer.Bytes fails to parse.

Do you have an example certificate for this test case?

chewong: Hmm most of the bad CA bundles I've tried were caught by `pem.Decode`. Still trying to look for a certificate that can pass `pem.Decode` but not `x509.ParseCertificate`

constraint/pkg/client/drivers/local/args.go
```diff
@@ -115,6 +120,23 @@ func DisableBuiltins(builtins ...string) Arg {
...
+	return func(d *Driver) error {
+		d.clientCertFile = certFile
+		d.clientKeyFile = keyFile
+
+		return nil
+	}
+}
+
+func EnableExternalDataClientAuth() Arg {
+	return func(d *Driver) error {
```
ritazh, line 133: why do you need to return error if it's always going to be nil?

maxsmythe: I think this is to fit the generic `Arg` interface?

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -321,6 +337,40 @@ func (d *Driver) Dump(ctx context.Context) (string, error) {
...
+
+	return ioutil.ReadAll(file)
+}
+
+func (d *Driver) getTLSCertificate() (*tls.Certificate, error) {
+	if !d.enableExternalDataClientAuth {
+		return nil, nil
+	}
+
+	certPEM, err := d.readFile(d.clientCertFile)
```
maxsmythe, line 356: IIRC you were going to look at the impact disk latency may have here given that we are reading this file from disk for every outbound request. What did you find?

constraint/pkg/externaldata/cache.go
```diff
@@ -60,15 +66,60 @@ func isValidName(name string) bool {
...
+func isValidCABundle(provider *v1alpha1.Provider) error {
+	// verify attempts to parse the caBundle as a PEM encoded certificate
+	// to make sure it is valid before adding it to the cache
+	verify := func(caBundle string) error {
+		caPem, err := base64.StdEncoding.DecodeString(caBundle)
+		if err != nil {
+			return fmt.Errorf("failed to base64 decode caBundle: %w", err)
+		}
+
+		caDer, _ := pem.Decode(caPem)
```
ritazh, line 85: do we also want to check for error here?

constraint/pkg/externaldata/cache.go
```diff
@@ -60,15 +66,60 @@ func isValidName(name string) bool {
...
+
+	u, err := url.Parse(provider.Spec.URL)
+	if err != nil {
+		return err
+	}
+
+	switch u.Scheme {
+	case HTTPScheme:
+		if !provider.Spec.InsecureTLSSkipVerify {
+			return fmt.Errorf("only HTTPS scheme is supported for Providers. To enable HTTP scheme, set insecureTLSSkipVerify to true")
```
ritazh, line 106:
```suggestion
			return fmt.Errorf("only HTTPS scheme is supported for this Provider. To enable HTTP scheme, set insecureTLSSkipVerify to true")
```

constraint/pkg/client/drivers/local/driver.go
chewong: I tested it on a non-SSD hard drive btw. It also depends on the type of disk the user is using, as you mentioned above. If reading from a spinning disk is too slow, they should consider upgrading it to a faster disk like an SSD.

constraint/pkg/externaldata/cache.go
chewong: I extracted this piece of code from https://github.com/open-policy-agent/cert-controller/blob/master/pkg/rotator/rotator.go#L524-L527.

ritazh (approved):
LGTM

chewong:
> pending failed gatekeeper tests

the gatekeeper test won't pass because this PR contains a minor breaking change (I changed the function signature of `SendRequestToProvider`)

constraint/pkg/client/drivers/local/driver.go
chewong: Done. PTAL.

PR #230, opened 2022-06-03
Add Constraint schema validation error

willbeason:
This is so that "gator verify" tests may check the schema of
Constraints. We need an error specific to schema validation so users can
explicitly test for this case.

Required for https://github.com/open-policy-agent/gatekeeper/issues/1989


sozercan (approved):
LGTM

maxsmythe (approved):
LGTM

PR #245, opened 2022-09-15
perf: Disable round trip on write

maxsmythe:
This disables the round-tripping of cached data between serialized/deserialized JSON when caching in OPA. This reduces the amount of time a lock is held on OPA's cache, which helps mitigate the lock contention mentioned in
https://github.com/open-policy-agent/gatekeeper/issues/2060


sozercan:
@maxsmythe looks like GK and unit tests are failing

maxsmythe:
Yeah, turns out RoundTrip had a different signature than I thought. Hopefully the newest change works.

constraint/go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/golang/glog v1.0.0
 	github.com/google/go-cmp v0.5.9
-	github.com/open-policy-agent/opa v0.42.2
+	github.com/open-policy-agent/opa v0.44.0
```
ritazh, line 21: This bump also fixes few CVEs. yay!

constraint/pkg/client/client.go
```diff
@@ -415,7 +416,22 @@ func (c *Client) AddData(ctx context.Context, data interface{}) (*types.Response
 			}
 		}
 
-		err = c.driver.AddData(ctx, name, key, processedData)
+		// Round trip data to force untyped JSON, as drivers are not type-aware
+		bytes, err := json.Marshal(processedData)
```
ritazh, line 420: For my own knowledge, why does doing round trip here improves performance compare to in storage?

maxsmythe: Storage acquires a mutex, so a round trip there blocks all Rego evaluation. Round trips here can be done in parallel without blocking Rego evaluation.

TL;DR less CPU time spent with a mutex

ritazh (approved):
LGTM

PR #248, opened 2022-09-16
fix: Enforce maximum constraint template name length of 63 chars

### LINKED ISSUE #116: ConstraintTemplate have no maximum name length
Constraint templates do not currently have a maximum name length.

It is necessary to cap the maximum possible constraint template name length to leave room for ConstraintTemplateStatus resources. Constraints have a maximum name length of 63 chars, we can be a bit more generous with Templates, if necessary.

Here is the PR that caps constraint name length, for reference:

https://github.com/open-policy-agent/frameworks/commit/b745ff7e8a941ade965188826050a98b62f151a2#diff-f0ceecee7c4c5529a7457604249a3c7e04bf92c547d1b42bf9af2e710cc4ea07

maxsmythe:
Gatekeeper breaks with longer template names.

Fixes #116


constraint/pkg/client/client_test.go
```diff
@@ -294,6 +294,15 @@ func TestClient_AddTemplate(t *testing.T) {
 			wantHandled: map[string]bool{handlertest.TargetName: true},
 			wantError:   nil,
 		},
+		{
+			name:    "Long name",
```
ritazh, line 298: You are adding a test here, where are you adding the validation?
> If at most 63 characters, then ConstraintTemplateNames can be up to 253-63=190 characters long.

maxsmythe: The validation is already happening on K8's side, since CRD kinds have a maximum size limit. This is just double checking that the check remains in place.

maxsmythe: i.e. the apiextension validation code we're calling performs this check

ritazh: I thought Constraints have a maximum name length of 63 chars, not constraint template. This test is throwing an error if the template name is greater than 63 chars?

ritazh: Will's comment [in the issue](https://github.com/open-policy-agent/frameworks/issues/116#issuecomment-1088829287) confused me
> If at most 63 characters, then ConstraintTemplateNames can be up to 253-63=190 characters long.

ritazh (approved):
lgtm

constraint/pkg/client/client_test.go
maxsmythe: yeah, I don't think we knew the validation code was being activated regardless

PR #252, opened 2022-10-01
perf: Remove `with` statement to avoid deepcopying data.inventory

maxsmythe:
This is a possible fix for https://github.com/open-policy-agent/gatekeeper/issues/2283

Digging through the OPA code, the `with` statement triggers a deepcopy of the referenced data. With the new execution model implemented for performance improvements, this deepcopy would be run multiple times, which could mean significant CPU overhead and unnecessary memory usage.

Not using the `with` key would make it harder to mask out only resources in specific namespaces if we ever create namespace-scoped constraint templates, but should not hurt near-term functionality.

It may be we will need an alternative method for accessing the data cache if we need to mask visibility in the future.

### DIFF
--- a/constraint/pkg/client/drivers/local/driver_unit_test.go
+++ b/constraint/pkg/client/drivers/local/driver_unit_test.go
@@ -768,7 +768,7 @@ func TestDriver_RemoveData_StorageErrors(t *testing.T) {
 			name: "commit error",
 			storage: &commitErrorStorage{
 				fakeStorage: fakeStorage{values: map[string]interface{}{
-					"/external/foo": "bar",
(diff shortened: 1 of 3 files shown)


maxsmythe:
@ritazh @sozercan We ran the numbers for `gator test` and this appears to eliminate the caching penalty that https://github.com/open-policy-agent/gatekeeper/pull/2301 was intended to address. I think this PR will benefit https://github.com/open-policy-agent/gatekeeper/issues/2256

sozercan (approved):
LGTM

ritazh (approved):

PR #257, opened 2022-11-08
update dependabot config to receive minor versions

sozercan:

so we can receive new OPA version bumps


maxsmythe (approved):
LGTM

PR #263, opened 2022-11-09
add EvaluationMeta empty interface to each Result

acpana:
### what

This patch adds an empty interface to keep evaluation metadata about a given `Result`.

### why

By collecting these measurements, we can expose them to consumers, like `gatekeeper` to use for reporting or logging.

#### reviewer notes:

- start with https://github.com/open-policy-agent/frameworks/pull/263/commits/4feb533a33b11b3ed358b076c651320f89025202

Signed-off-by: Alex Pana <8968914+acpana@users.noreply.github.com>


constraint/pkg/client/client.go
```diff
@@ -553,9 +555,14 @@ func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.Qu
...
 		}
 
+		// edge case: .Milliseconds() can be rounded to 0 so take the nanos and convert to millis
+		resp.EvaluationLatency = float64(time.Since(reviewStart).Nanoseconds()) / 1000000
+
 		for _, autorejection := range autorejections[target] {
 			resp.AddResult(autorejection.ToResult())
 		}
+		// since autorejections don't get evaluated, adjust the batchSize to include those
+		resp.BatchSize += uint(len(autorejections))
```
acpana, line 565: At first, I didn't add this but it makes sense to add the auto rejections to the `BatchSize` for the following reasons:

- accurately represent the `Review` objects that were actually under the `.Review()` call

constraint/pkg/types/validation.go
```diff
@@ -9,6 +9,27 @@ import (
...
 )
 
+type EvaluationEngineType string
+
+const (
+	Undefined  EvaluationEngineType = "undefined"
+	RegoEngine EvaluationEngineType = "rego"
+)
+
+type ReviewMeta struct {
```
acpana, line 19: I like the extra level of indirection here for the following reasons:

* we may add different `metadata` things to a `ResponseMeta` and keep things somewhat modular
* we could reuse `ReviewMeta` if we wanted to add such a thing to the `Result` type itself.

acpana:
@maxsmythe let me know what you think of the open questions above! 🙏🏼

constraint/pkg/types/validation.go
acpana: Expanding on this a bit:

> * we may add different `metadata` types to a `ResponseMeta` and keep the specific logic somewhat modular

One small papercut here is that `Response` can be returned via multiple API calls. We mostly care about the `ReviewMeta` from the `.Review()` call; if we ever need more API/ operation specific metadata, we could expose those as needed (ie `AddTemplateMeta`, etc.).

constraint/pkg/types/validation.go
```diff
@@ -9,6 +9,27 @@ import (
...
+	// EvaluationLatency is the number of milliseconds it took to server a client.Review() call.
+	EvaluationLatency float64
```
acpana, line 21: At first glance, latency and batch size can be easily determined by a consumer of the API call, like the gatekeeper at the call site.

But the underlying measurement here is the evaluation time during the engine call (`driver.Query()`)., which is what we are trying to surface. This way, future iterations of the `Review` API have a notion of the time spent under evaluation for a `review` object and not how long the `Review()` func took to work/ execute overall.

constraint/pkg/client/client.go
```diff
@@ -553,9 +555,14 @@ func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.Qu
 			continue
 		}
 
+		// edge case: .Milliseconds() can be rounded to 0 so take the nanos and convert to millis
+		resp.EvaluationLatency = float64(time.Since(reviewStart).Nanoseconds()) / 1000000
```
acpana, line 559: Notice we are just collecting this information. Consumers of the library will log/ emit/ analyze this as needed.

constraint/pkg/client/client.go
```diff
@@ -588,6 +595,13 @@ func (c *Client) review(ctx context.Context, target string, constraints []*unstr
 		Trace:   trace,
 		Target:  target,
 		Results: results,
+		ResponseMeta: &types.ResponseMeta{
```
maxsmythe, line 598: The intended use case for this feature would be: if I have 5 constraint templates and one of them is slow, I would like to identify which template is the culprit.

In order to do that, I'd need to be able to associate a given constraint template with its total runtime.

TL;DR we need to break down these `Meta` objects by constraint template name... probably by modifying the engine interface to allow that.

constraint/pkg/types/validation.go
```diff
@@ -9,6 +9,27 @@ import (
...
+type EvaluationEngineType string
+
+const (
+	Undefined  EvaluationEngineType = "undefined"
+	RegoEngine EvaluationEngineType = "rego"
+)
+
+type ReviewMeta struct {
+	// evaluationLatency is the number of milliseconds it took to server a client.Review() call.
+	evaluationLatency float64
```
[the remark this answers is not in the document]
acpana: I think `TotalTemplateRunTime` can box us into not easily using the `ReviewMeta` to show the run time per constraint.

`evaluationRunTime` ?

constraint/pkg/types/validation.go
```diff
@@ -23,6 +44,8 @@ type Result struct {
 
 	// The enforcement action of the constraint
 	EnforcementAction string `json:"enforcementAction,omitempty"`
+
+	*ResultMeta `json:"-,inline"`
```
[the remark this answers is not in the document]
acpana: I tried to touch on the level of double indirection in a comment before.

tl;dr- we can use `ResultMeta` to stuff other metadata information related to the `Result`, other than `ReviewMeta`.

constraint/pkg/types/validation.go
```diff
@@ -9,6 +9,27 @@ import (
...
+type ReviewMeta struct {
+	// evaluationLatency is the number of milliseconds it took to server a client.Review() call.
+	evaluationLatency float64
+	// engineType denotes an enum for which kind of underlying engine was used for a client.Review() call.
+	engineType EvaluationEngineType
+	// batchSize indicates how many constrains were evaluated for an underlying engine eval call.
+	batchSize uint
+}
+
+// ResultMeta coontains metadata, such as latency, etc., for a given Response.
```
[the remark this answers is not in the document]
acpana: I tried to touch on the level of double indirection in a comment before.

tl;dr- we can use `ResultMeta` to stuff other metadata information related to the `Result`, other than `ReviewMeta`. 

wdyt?

constraint/pkg/types/validation.go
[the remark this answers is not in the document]
maxsmythe: The idea behind `GetStatsString()` is that the reporting is amorphous, which frees us from needing to figure out these details.

That aside `TotalTemplateRunTime` would actually free us up to report per-constraint metrics later. I'm assuming you are choosing the name `evaluationRunTime` because the scope is ambiguous, which would leave you room to change the meaning of the number later. Thing is, that would avoid breaking type checking for users who relied on that code, but change the meaning of the value, which could break the assumptions of whatever is built on top of it.

TL;DR:

* We should consider GetStatsString() the interface, and probably make these other members private
* Backwards compatibility is as much about the meaning of a field staying content as its name

constraint/pkg/types/validation.go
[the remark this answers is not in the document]
maxsmythe: I think using interface types for a variable moots this.

Also, if the name is salient, I probably wouldn't use embedded structs, those are more for composition-type design (e.g. how K8s uses ObjectMeta)

constraint/pkg/types/validation.go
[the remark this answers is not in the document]
acpana: new `ResultMeta` has no json struct fields on it anymore

acpana:
hey @maxsmythe thanks for the guidance here. 

Starting with [8057241](https://github.com/open-policy-agent/frameworks/pull/263/commits/8057241a3c3e76e68440a51bdca6597769438748), we have an interface exposing an `EngineStatsString` ;

constraint/pkg/types/validation.go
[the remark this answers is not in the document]
acpana: Thanks for the input here! I reworked the code per the guidance.

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -72,6 +73,20 @@ type Driver struct {
 	clientCertWatcher *certwatcher.CertWatcher
 }
 
+// regoResultMeta has rego specific metadata for a result.
+type regoResultMeta struct {
+	// templateRunTime is the number of milliseconds it took to evaluate all constraints for a template.
+	templateRunTime float64
+	// engineType is an engine set string for the kind of underlying engine that was used for a client.Review() call.
```
maxsmythe, line 80: Given that this is `regoResultMeta`, we can assume engineType will always be `Rego`, right?

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -72,6 +73,20 @@ type Driver struct {
...
+	engineType string
+	// constraintCount indicates how many constraints were evaluated for an underlying engine eval call.
+	constraintCount uint
+}
+
+func (rm *regoResultMeta) EngineStatsString() string {
+	return fmt.Sprintf("totalTemplateRuntime: %.4f, engineType: %s, constraintCount: %d", rm.templateRunTime, rm.engineType, rm.constraintCount)
```
maxsmythe, line 87: If engine type is important for analysis, we may want to have that be reported by the framework to guarantee uniformity of how its reported. For now it can be omitted.

constraint/pkg/types/validation.go
```diff
@@ -134,3 +136,13 @@ func (r *Responses) TraceDump() string {
...
 }
+
+// ResultMeta defines an interface to expose metadata for a Result.
+type ResultMeta interface {
+	// EngineStatsString gives an amorphous representation of engine stats
+	// such as latency, engine type or the number of constraints evalauted
+	// against a template. The underlying enigne type defines and implements
+	// structs for this interface. This function is primarily inteded for logging
+	// out for the consumers of this library.
+	EngineStatsString() string
```
maxsmythe, line 147: no need for `String`

Another option here would be to return map[string]interface{}, which would make it play a bit more nicely with other serialization pipelines (such as the logs pipeline), and allow for programatic inspection should someone want to try.

acpana: I changed the return type to `map[string]interface{]` but also the signature to `ditto , error`; where an error occurs on json.marshalling/ unmarshalling

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -72,6 +73,14 @@ type Driver struct {
 	clientCertWatcher *certwatcher.CertWatcher
 }
 
+// RegoEvaluationMeta has rego specific metadata from evaluation.
+type RegoEvaluationMeta struct {
+	// TemplateRunTime is the number of milliseconds it took to evaluate all constraints for a template.
+	TemplateRunTime float64 `json:"templateRunTime"`
```
maxsmythe, line 79: should we add the units to this field?

maxsmythe (approved):
LGTM with one question around whether we want to bake units into the field name for clarity. Thanks for this!

acpana:
AFAIU, the `gatekeeper` test wf here will fail since `EvaluationMeta` won't be present. Is that ok to ignore (barring no other failures) in order to merge and then update the gatekeeper tests?

Here's an old gatekeeper test run: https://github.com/open-policy-agent/frameworks/actions/runs/3586143965/jobs/6037544599 (it seems mostly gator test related changes needed)

maxsmythe:
This shouldn't be a breaking change, I'd expect Gatekeeper Test to pass, that being said, that test is informational

maxsmythe:
Looking at the failure you linked, it seems like some of G8r's tests are written in a brittle fashion, such that the addition of a new field causes them to break.

These are benign, though it does mean updating the dependency will take some work.

constraint/pkg/client/drivers/local/driver.go
maxsmythe: K8s API conventions prefer integrating them into the field:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

sozercan (approved):
LGTM

PR #270, opened 2022-12-07
feat: add externaldata v1beta1 api

sozercan:

GK PR: https://github.com/open-policy-agent/gatekeeper/pull/2438


constraint/pkg/externaldata/request.go
```diff
@@ -57,10 +57,10 @@ func NewProviderRequest(keys []string) *ProviderRequest {
 }
 
 // SendRequestToProvider is a function that sends a request to the external data provider.
-type SendRequestToProvider func(ctx context.Context, provider *v1alpha1.Provider, keys []string, clientCert *tls.Certificate) (*ProviderResponse, int, error)
+type SendRequestToProvider func(ctx context.Context, provider *v1beta1.Provider, keys []string, clientCert *tls.Certificate) (*ProviderResponse, int, error)
```
maxsmythe, line 60: use unversioned here?

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -54,10 +57,44 @@ spec:
...
+            type: object
+          spec:
+            description: Spec defines the Provider specifications.
+            properties:
+              caBundle:
+                description: CABundle is a base64-encoded string that contains the
+                  TLS CA bundle in PEM format. It is used to verify the signature
+                  of the provider's certificate.
+                type: string
+              insecureTLSSkipVerify:
```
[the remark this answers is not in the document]
sozercan: we can still require `https` and deprecate this field.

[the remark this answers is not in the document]
maxsmythe: We actually can't per K8s backwards compatibility guidelines:

This would be a behavioral change:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

And v1beta1 signals backwards compatibility:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#incompatible-api-changes

After we move to v1beta1, we can no longer deprecate the field without risking breaking users who depend on it. I can see exceptions for unforeseen circumstances or security vulnerabilities, but that's not the case here.

TL;DR once we go to beta, if this field exists, it's here to stay. Personally, I'm fine with that, but it's worth highlighting.

constraint/pkg/externaldata/cache.go
```diff
@@ -8,32 +8,32 @@ import (
...
 type ProviderCache struct {
-	cache map[string]v1alpha1.Provider
+	cache map[string]v1beta1.Provider
 	mux   sync.RWMutex
 }
 
 func NewCache() *ProviderCache {
 	return &ProviderCache{
-		cache: make(map[string]v1alpha1.Provider),
+		cache: make(map[string]v1beta1.Provider),
```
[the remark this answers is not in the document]
sozercan: updated to unversioned

constraint/pkg/externaldata/request.go
sozercan: updated to unversioned

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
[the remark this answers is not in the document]
enj: Why not simply drop this field?  Require `https` and support `http` on `::1` and `127.0.0.1`.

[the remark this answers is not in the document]
sozercan: Removed `insecureTLSSkipVerify` for now.

maxsmythe (approved):
LGTM!

constraint/pkg/externaldata/cache.go
```diff
@@ -101,15 +101,6 @@ func isValidCABundle(provider *unversioned.Provider) error {
...
-			return fmt.Errorf("only HTTPS scheme is supported for this Provider. To enable HTTP scheme, set insecureTLSSkipVerify to true")
-		}
-		if provider.Spec.CABundle != "" {
-			if err := verify(provider.Spec.CABundle); err != nil {
-				return err
-			}
-		}
 	case HTTPSScheme:
 		if provider.Spec.CABundle == "" {
 			return fmt.Errorf("caBundle should be set for HTTPS scheme")
```
enj, line 106: I do not understand why this is a requirement.  Normally when making a TLS connection with no pre-configured CA bundle, the system verifier is used (by leaving `tls.Config.RootCAs` as `nil`).

sozercan: let's handle this in a follow up. Opened #2452 for tracking

ritazh:
Why is gatekeeper_test broken?

sozercan:
> Why is gatekeeper_test broken?

Supposed to be broken. https://github.com/open-policy-agent/gatekeeper/pull/2438 addresses those

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -55,9 +54,39 @@ spec:
...
+                description: Timeout is the timeout when querying the provider.
+                type: integer
+              url:
+                description: URL is the url for the provider. URL is prefixed with
+                  http:// or https://.
+                type: string
+            type: object
+        type: object
+    served: true
+    storage: false
```
ritazh, line 92: shouldn't storage be true here and false for v1alpha1?

sozercan: I discussed this with @maxsmythe. for backwards compat, we thought about leaving it in v1alpha1 for now

constraint/pkg/externaldata/cache_test.go
```diff
@@ -154,7 +153,7 @@ func TestRemove(t *testing.T) {
 	tc := []cacheTestCase{
 		{
 			Name:          "valid provider",
-			Provider:      createProvider("test", "http://test", 1, "", true),
+			Provider:      createProvider("test", "http://test", 1, ""),
```
ritazh, line 156: should http be valid?

constraint/pkg/externaldata/request.go
```diff
@@ -100,44 +100,43 @@ func DefaultSendRequestToProvider(ctx context.Context, provider *v1alpha1.Provid
...
+
 	client := &http.Client{
 		Timeout: time.Duration(provider.Spec.Timeout) * time.Second,
 	}
 
-	tlsConfig := &tls.Config{
-		//nolint:gosec
-		InsecureSkipVerify: provider.Spec.InsecureTLSSkipVerify,
-	}
+	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS13}
```
ritazh, line 118: Should min TLS version be configurable? we should at least add this to the docs.

sozercan: I think if we are making this secure by default. It's okay to set min as 1.3 to match GK. I'll add to docs.

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
ritazh: Can you please explain or link me to the rationale for this?

constraint/pkg/externaldata/cache_test.go
sozercan: no http is not valid, https only

sozercan: good catch, this is a remove test. I'll update to https

constraint/pkg/externaldata/request.go
enj: For new integrations, there is no reason to not use TLS 1.3.  TLS 1.2 requires careful cipher control to get a secure config.  It is [a lot of work](https://github.com/vmware-tanzu/pinniped/blob/main/internal/crypto/ptls/default.go).

sozercan: added to docs: [open-policy-agent/gatekeeper@`767cd9e` (#2438)](https://github.com/open-policy-agent/gatekeeper/pull/2438/commits/767cd9e9ca952160e0f54dccb4562694efabe7fc)

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
maxsmythe: If a user rolls back G8r (including the ExternalData CRD), the stored data can continue to be served otherwise, a CRD that only knows v1alpha1 may break if it reads a different storage version. No other benefit AFAIK.

ritazh: Gotcha. When is it safe for us to change the beta storage to true? I don't recall doing this for other GK CRDs moving from alpha to beta. Should we follow the same suit in the future when we update the API for all GK CRDs? Or is the ExternalData CRD unique?

maxsmythe: This is the first time we've done it. Probably not a bad idea to keep doing it moving forward.

As to when we'd shift... I guess when we have one stable version out that has v1beta1 in it so rolling back the CRD doesn't break things?

ritazh: Sounds good! created https://github.com/open-policy-agent/gatekeeper/issues/2457 to track it for v3.12

sozercan:
@ritazh good to merge?

ritazh (approved):
LGTM

PR #276, opened 2023-01-18
feat: backfill nil results after engine evaluation

acpana:
TODO/ wip

This patch contains two things:

* renames a field to include units: `TemplateRunTimeMillis`

and 

* adds an option to backfill results after the engine evaluation. This can be helpful to show the template run time in gatekeeper for objects under test that don't cause a violation for a constraint, but would still take time to evaluate. This in turn helps with debugging the question "which rule is slow?"

The change is not cost free imo. The extra work of backfilling and send that over to gatekeeper takes compute time itself. More importantly, there are places in the gatekeeper code base that assume that any result returned from `Review` is a violation. Integrating this change in g8r will require taking a pass over those assumptions.

Note:
* the patch has been tested locally w a replace directive in g8r to show that the `EvaluationMeta` show up and show up for objects that don't violate a constraint.

### DIFF
--- a/constraint/pkg/client/drivers/local/args.go
+++ b/constraint/pkg/client/drivers/local/args.go
@@ -68,6 +68,14 @@ func PrintEnabled(enabled bool) Arg {
 	}
 }
 
+func BackfillResults(enabled bool) Arg {
+	return func(d *Driver) error {
+		d.backfillResults = enabled
+
+		return nil
+	}
+}
+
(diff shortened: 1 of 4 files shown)


acpana:
This was not a great idea given that we'd need a few changes to Gatekeeper and that we are inflating some of the concepts here.

Closing in favor of a more generalized approach to reporting the metadata: https://github.com/open-policy-agent/frameworks/pull/283

PR #279, opened 2023-01-19
ci: remove kubebuilder dependency

sozercan:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -22,7 +22,7 @@ jobs:
   test:
     name: "Unit test"
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 10
     steps:
       - name: Set up Go 1.19
         uses: actions/setup-go@v3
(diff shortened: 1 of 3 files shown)


sozercan:
GK test should succeed after https://github.com/open-policy-agent/gatekeeper/pull/2524 is merged

.github/workflows/workflow.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   test:
     name: "Unit test"
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 10
```
acpana, line 25: hm, did you notice an increase in timeouts?

sozercan: yes, i added `-race -bench` to align with GK which increased the timeouts

sozercan: the reason we added is to make sure to run benchmarks often so we don't break them

constraint/Makefile
```diff
@@ -109,3 +107,22 @@ generate-defaults: constraint-template-string-constant
...
+
+.PHONY: __test-image
+__test-image:
+	docker buildx build . \
+		-t constraint-test \
+		--load \
+		--build-arg KUSTOMIZE_VERSION=$(KUSTOMIZE_VERSION)
+
+## Location to install dependencies to
+LOCALBIN ?= $(shell pwd)/bin
```
maxsmythe, line 119: If a user runs this in their source code directory, this will create a /bin folder. Do we want to put this in an easy-to-clean subdir (.output or similar)? Is the /bin directory part of .gitignore?

maxsmythe: looks like bin would be ignored, though I'd personally consider putting this in a .output directory to make it easy to wipe everything and start from zero.

maxsmythe (approved):
LGTM

constraint/Makefile
sozercan: @maxsmythe `bin` is the default for kubebuilder v3. looks like we are using `.tmp` already (for linter) so i changed to `.tmp/bin`

maxsmythe (approved):
Sorry, I thought I LGTMd this after verifying bin was git-ignored (meant that to be a nit)

maxsmythe:
It looks like the tests are now failing. If moving the binaries makes it harder to start the control plane, let's just use the vanilla location.

PR #283, opened 2023-01-27
feat: add more generalized Stats

acpana:
Signed-off-by: Alex Pana <8968914+acpana@users.noreply.github.com>

This patch reworks `EvaluationMeta` field as a more general struct.

The big win here is seeing `Stats` for non violating constraints
without manipulating the `Result` struct, which is not possible with the 
implementation in https://github.com/open-policy-agent/frameworks/pull/263. This was evident when we tried to integrate in gatekeeper.

#### current papercuts
You will notice we modify the  driver interface for `Query()` to give back a `QueryResponse` type. This breaks compatibility for drivers. We need to modify the `Query()` signature regardless to give back stateless, per call metrics. Adding the `QueryResponse` struct helps future proof for further additions.

#### UX in action
Here's what consuming this, say in g8r, would look like w `fmt.Println`:

```bash
scope: template, statsFor: K8sGoodRego, source: rego, statsName: templateRunTimeNanos, value: 205970, desc: the number of milliseconds it took to evaluate all constraints for a template 
scope: template, statsFor: K8sGoodRego, source: rego, statsName: constraintCount, value: 1, desc: the number of constraints that were evaluated at the same time for the given constraint kind 
```

These can be logged or sent to a metrics aggregator depending on needs.

#### call outs
* note that since we are changing the `Result` struct, some tests in gatekeeper are failing and that's ok


constraint/pkg/client/drivers/interface.go
```diff
@@ -45,6 +46,11 @@ type Driver interface {
 	// Dump outputs the entire state of compiled Templates, added Constraints, and
 	// cached data used for referential Constraints.
 	Dump(ctx context.Context) (string, error)
+
+	// TODO docs
+	EvaluationStats() string
+
+	StatsEntries() []*instrumentation.StatsEntry
```
acpana, line 53: NOTE:

These two funcs server the same purpose: `exposing stats out of the driver`

I want us to pick one of the two and I am more partial to `StatsEntries` as it returns data in a more structured way and leaves it up to the consumers to define the format of emission (logging, otel, etc.)

That said, the previous implementation had even less structure to the stats and just exposed a string out of the driver.

constraint/pkg/client/drivers/local/instrumentation.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return b.String()
+}
+
+func (m *metricsReporterImpl) StatsEntries() []*instrumentation.StatsEntry {
+	if m.store == nil {
+		return nil
+	}
+
+	var entries []*instrumentation.StatsEntry
```
davis-haba, line 37: nit:
`entries := make([]*instrumentation.StatsEntry, len(m.store))`
might be more efficient. If the slice isn't going to grow: `make([]*instrumentation.StatsEntry, len(m.store), len(m.store))

constraint/pkg/client/drivers/interface.go
acpana: Given the above and chatting internally too, I really like the UX of sending over more structured data. So the `string` approach is going away in the latest commit.

constraint/pkg/client/drivers/local/instrumentation.go
acpana: n/a know after `4cdb922` bc we are not using a Reporter anymore

constraint/pkg/client/client.go
```diff
@@ -584,10 +586,17 @@ func (c *Client) review(ctx context.Context, target string, constraints []*unstr
 		tracesBuilder.WriteString("\n\n")
 	}
 
+	var statsEntries []*instrumentation.StatsEntry
+	if localDriver, ok := c.driver.(*local.Driver); ok {
```
acpana, line 590: if we're fine w this approach, we could make the `local.Driver` have a func of its own. I'd just be weary about modifying the interface w the current approach (of storing stats in a variable) to not back ourselves in a corner. That is, other drivers can have a different style of storing and handling `StatsEntries`

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -71,14 +73,13 @@ type Driver struct {
...
-	// TemplateRunTime is the number of milliseconds it took to evaluate all constraints for a template.
-	TemplateRunTime float64 `json:"templateRunTime"`
-	// ConstraintCount indicates how many constraints were evaluated for an underlying rego engine eval call.
-	ConstraintCount uint `json:"constraintCount"`
+	// gatherMetrics controls whether the driver attempts to gather any metrics around its API calls
+	gatherMetrics bool
+
+	// StatsEntries returns the current statistics observed since the last API call. At present
+	// StatsEntries are only gathered on the Query() API.
+	StatsEntries []*instrumentation.StatsEntry
```
acpana, line 82: We could make this unexported but then we'd need some brittle reflection to access it. One pro of going that way would be that CF users wouldn't be tempted to use this.

constraint/pkg/client/drivers/local/driver_unit_test.go
```diff
@@ -171,18 +181,179 @@ func TestDriver_Query(t *testing.T) {
 	if len(res) == 0 {
 		t.Fatalf("got 0 errors on data-less query; want 1")
 	}
+}
 
-	stats, ok := res[0].EvaluationMeta.(RegoEvaluationMeta)
-	if !ok {
-		t.Fatalf("could not type convert to RegoEvaluationMeta")
+// TestDriver_Query_StatsEntries tests that StatsEntries are populated for Query calls.
```
acpana, line 186: I went w a longer test, split test here for readability. IMO, there are not that many cases to test here and a table test may be confusing. If we are adding more and more cases, then I can be persuaded to convert these.

acpana (commented):
Adding some context;

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -290,14 +314,37 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
+						{
+							StatsName: "EvaluationMetadata",
+							Stats: []*instrumentation.Stat{
+								{
+									Name:        "templateRunTimeMillis",
+									Value:       fmt.Sprintf(" %.2f", evalRunTime),
+									Description: "the number of milliseconds it took to evaluate all constraints for a template",
+								},
+								{
+									Name:        "engineType",
```
maxsmythe, line 336: I like the idea of reporting the engine type. This is an example of structure that we can enforce by being intentional about how we collate the responses to multiple calls to Query() (likely in client.go)

constraint/pkg/client/drivers/local/driver_unit_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"reflect"
```
maxsmythe, line 10: Let's get the basic pattern down before writing tests. Changes in architecture will likely cause these tests to be thrown out.

Also, if we come up with a simpler model, the tests should be simpler to write.

If the concern is Test Driven Development... usually that requires knowledge of what the interface looks like, which is what we're trying to define.

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,24 @@
...
+	StatsName string
+	Stats     []*Stat
+}
+
+// Stat is a Name, Value, Description tuple that may contain
+// metrics or aggregations of metrics.
+type Stat struct {
+	Name        string
+	Value       string
+	Description string
```
maxsmythe, line 17: Description could be valuable for giving drivers a way to document their responses

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,24 @@
...
+// to be represented to consumers.
+type NamedStats struct {
+	StatsName string
+	Stats     []*Stat
+}
+
+// Stat is a Name, Value, Description tuple that may contain
+// metrics or aggregations of metrics.
+type Stat struct {
+	Name        string
```
maxsmythe, line 15: how is `Name` different from `StatsName`?

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,24 @@
+// package instrumentation defines primitives to
+// support more observability throughout the contraint framework.
+package instrumentation
+
+// NamedStats covers primitive for many heterogenoues statstics and metrics
+// to be represented to consumers.
+type NamedStats struct {
+	StatsName string
+	Stats     []*Stat
```
maxsmythe, line 9: What is an example use case that would leverage a list of stats here?

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// Stat is a Name, Value, Description tuple that may contain
+// metrics or aggregations of metrics.
+type Stat struct {
+	Name        string
+	Value       string
```
maxsmythe, line 16: Will Value always be string?

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,24 @@
...
+	Description string
+}
+
+// StatsEntry comprises of a generalized Key for all associated Stats.
+type StatsEntry struct {
+	Key      interface{}   `json:"key"`
```
maxsmythe, line 22: What is `key` intended for?

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -224,6 +240,17 @@ func (d *Driver) eval(ctx context.Context, compiler *ast.Compiler, target string
 }
 
 func (d *Driver) Query(ctx context.Context, target string, constraints []*unstructured.Unstructured, review interface{}, opts ...drivers.QueryOpt) ([]*types.Result, *string, error) {
+	if d.gatherMetrics {
+		d.mtx.Lock()
```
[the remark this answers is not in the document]
acpana: Right, that's why we'd only write lock if we are gathering metrics in the driver. I also called out poor performance of the current approach in the PR notes.

constraint/pkg/instrumentation/instrumentation.go
acpana: the current use case ( `evaluation metadata` / `perf metrics` showcase this:

-we have the name of all the stats gathered under `EvaluationMetadata`; `Name` here is the name of the collection of `Stat` structs
-two `Stat`s structs:
--`templateRunTimeMillies` & `engineType` each recording their own `Value` and `Description` & `Name`; `Name` here is the name of the `Stat` 

does this help? If so, I can add the explanation in as comments.

constraint/pkg/instrumentation/instrumentation.go
acpana: see above https://github.com/open-policy-agent/frameworks/pull/283/files#r1093610950

constraint/pkg/instrumentation/instrumentation.go
acpana: i was thinking so. At first, I thought we could have it as a more generic type but decided to start w string since consumers are free to take on the conversion

constraint/pkg/instrumentation/instrumentation.go
acpana: the `Key` here represents what struct/ object the `NamedStats` list are for.

In our case, we probably will have `template` names but down the road the `key` could be `ConstraintKey` or a `Constraint` name

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -71,14 +73,28 @@ type Driver struct {
...
 
-// RegoEvaluationMeta has rego specific metadata from evaluation.
-type RegoEvaluationMeta struct {
-	// TemplateRunTime is the number of milliseconds it took to evaluate all constraints for a template.
-	TemplateRunTime float64 `json:"templateRunTime"`
-	// ConstraintCount indicates how many constraints were evaluated for an underlying rego engine eval call.
-	ConstraintCount uint `json:"constraintCount"`
+// StatsEntries returns any stats associated with the driver API if GatherMetrics
+// driver arg was passed in. The only current endpoinut supported is Query().
+func (d *Driver) StatsEntries() []*instrumentation.StatsEntry {
```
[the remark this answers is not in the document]
acpana: If we must return the stats as the same time as the current `Query()` calls then this could break the backwards compatibility of the interface since we'd have to modify the return types.

We could add a `QueryWStats()` (or some better named func) as previously discussed that returns stats (in addition to what `Query` returns too).

I'm partial to the `QueryWStats()` avenue since we wouldn't necessarily have to modify the `Driver` interface to accommodate that func on the local driver and keep casting the driver in the client. As w iterate on Stats & the Driver this could help us shield the interface form multiple changes. Wdyt?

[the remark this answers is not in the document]
maxsmythe: No need for an extra function... this is `v0.x.x` for a reason.

Also, I don't think I agree with you that backwards compatibility necessarily needs to be broken, the mere altering of a struct doesn't break compatibility, it depends on how it's altered. See K8s backwards compatibility docs for some examples:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

However, we __know__ we are breaking backwards compatibility in the sense that we are removing the previous stats work.

constraint/pkg/client/drivers/local/driver.go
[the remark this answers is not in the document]
maxsmythe: This functionality should not require a write lock at all.

constraint/pkg/instrumentation/instrumentation.go
maxsmythe: Why would the collection of stat structs have a name? What would be an example name?

constraint/pkg/instrumentation/instrumentation.go
maxsmythe: Unfortunately that doesn't answer the question as-written, but we can continue the discussion there.

constraint/pkg/instrumentation/instrumentation.go
maxsmythe: Conversion in the sense of parsing strings can be expensive and error-prone. If we want to preserve the potential for strong types, some kind of `interface{}` value is likely a better fit (maybe require it supports a String() method? Worth looking at what would be necessary to make things easy to JSON-ify).

constraint/pkg/instrumentation/instrumentation.go
maxsmythe: Ah, `key` is probably not descriptive for this use case... something like `kind` or `scope` would be better, though I'm sure there are even more fitting names.

This is in support of some kind of aggregation?

constraint/pkg/client/drivers/local/args.go
```diff
@@ -154,6 +155,17 @@ func Externs(externs ...string) Arg {
 	}
 }
 
+// GatherMetrics starts collecting the various metrics around the
+// underlying engine's eval() calls.
+func GatherMetrics() Arg {
+	return func(driver *Driver) error {
+		driver.gatherMetrics = true
```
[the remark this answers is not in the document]
acpana: the latest approach moves away from storing the metrics in an inmem ds

constraint/pkg/instrumentation/instrumentation.go
acpana: `NamedStats` have been removed per our offline chat.

constraint/pkg/instrumentation/instrumentation.go
acpana: We removed `NamedStats`

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -287,25 +296,45 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
-				ConstraintCount: uint(len(kindResults)),
-			}
+			return &drivers.QueryResponse{Err: err}
 		}
 
 		results = append(results, kindResults...)
+
+		if d.gatherMetrics {
+			// for the batch of evaluated constraints, report their evaluation metadata
+			evalRunTime := float64(evalEndTime.Nanoseconds()) / 1000000
```
maxsmythe, line 306: What is the impact on serialization/deserialization of float vs. int?

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -287,25 +296,45 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
+
+			statsEntries = append(statsEntries,
+				&instrumentation.StatsEntry{
+					Key: kind,
+					Stats: []*instrumentation.Stat{
+						{
+							Name:        templateRunTimeMillisName,
```
maxsmythe, line 313: For multi-engine, knowing what engine created these metrics could also be useful

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -287,25 +296,45 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
+							Value:       &instrumentation.Float64Value{ConcreteValue: evalRunTime},
+							Description: templateRunTimeMillisDescription,
```
maxsmythe, line 315: How do we know which constraint template this metric is associated with?

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// StatsEntry comprises of a generalized Key for all associated Stats.
+type StatsEntry struct {
+	Key   interface{} `json:"key"`
+	Stats []*Stat     `json:"stats"`
+}
+
+type Value interface {
+	String() string
```
maxsmythe, line 26: IIRC all golang types support String() one way or another (I may be wrong about this), but generally are string-ifiable. What is the benefit of Value() over just `interface{}`?

maxsmythe (commented):
I like the aggregation model that's starting to take shape here.

constraint/pkg/instrumentation/instrumentation.go
acpana: I think the main selling point for me, to my mind, was to trim the float64 up to 2 decimal points. I can also buy the argument to leave that up to the consumer after a type cast.

constraint/pkg/client/drivers/local/driver.go
acpana: are you asking in terms of performance?

constraint/pkg/client/drivers/local/driver.go
acpana: +1 to that! We are capturing the engine type lower here on line 318

constraint/pkg/client/drivers/local/driver.go
acpana: the general shape of `StatsEntry` is 

```
Key interface{}
Stats []*Stats
```

`Key ` here is specified on line 310 and it refers to the constraint kind evaluated. Here's what a `StatsEntry` would look like marshalled

```
    "key": "K8sUniqueServiceSelector",
    "stats": [
      {
        "name": "templateRunTimeMillis",
        "value": 0.37196,
        "description": "the number of milliseconds it took to evaluate all constraints for a template"
      },
      {
        "name": "engineType",
        "value": "rego",
        "description": "the underlying engine that performed the evaluation"
      },
      {
        "name": "constraintCount",
        "value": 1,
        "description": "the number of constraints that were evaluated at the same time for the given constraint kind"
      }
    ]
```

constraint/pkg/instrumentation/instrumentation.go
acpana: I thought about it some more and I think we can leave it up to consumers to decide how to typecast; I think it's cleaner from the CF end.

constraint/pkg/client/drivers/local/driver.go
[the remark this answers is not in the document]
acpana: Per this comment and offline chats, we decided to modify the `Query` func on the driver interface to return a `QueryResponse`. We'd embed the `StatsEntries` in that response.

constraint/pkg/client/drivers/local/driver.go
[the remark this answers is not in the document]
acpana: n/a after re-write

constraint/pkg/client/drivers/local/driver.go
acpana: this was a good call out! I was focused more on providing data for every `kind, name` tuple for constraints per the result backfilling mindset. Reworked this for `kind` level stats only.

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -290,14 +314,37 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
-			result.EvaluationMeta = RegoEvaluationMeta{
-				TemplateRunTime: float64(evalEndTime.Nanoseconds()) / 1000000,
-				ConstraintCount: uint(len(kindResults)),
+		results = append(results, kindResults...)
+
+		if d.gatherMetrics {
+			// for the batch of evaluated constraints, report their evaluation metadata
+			evalRunTime := float64(evalEndTime.Nanoseconds()) / 1000000
+
+			for _, k := range kindConstraints {
```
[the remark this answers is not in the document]
acpana: this was a good call out! I was focused more on providing data for every `kind, name` tuple for constraints per the result backfilling mindset. Reworked this for `kind` level stats only.

constraint/pkg/instrumentation/instrumentation.go
acpana: Renamed to `Scope`. Per our offline sync up, `Scope` can be the template name or the constraint name (down the road) or some other, more complex key for grouping the stats.

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -224,6 +240,17 @@ func (d *Driver) eval(ctx context.Context, compiler *ast.Compiler, target string
 }
 
 func (d *Driver) Query(ctx context.Context, target string, constraints []*unstructured.Unstructured, review interface{}, opts ...drivers.QueryOpt) ([]*types.Result, *string, error) {
+	if d.gatherMetrics {
+		d.mtx.Lock()
+		defer d.mtx.Unlock()
+
+		// reset stats entries every time
+		d.statsEntries = make([]*instrumentation.StatsEntry, 0)
```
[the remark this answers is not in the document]
acpana: n/a after rewrite

constraint/pkg/client/drivers/local/driver.go
```diff
@@ -30,6 +32,16 @@ import (
 const (
 	libRoot   = "data.lib"
 	violation = "violation"
+
+	templateRunTimeMillisName        = "templateRunTimeMillis"
+	templateRunTimeMillisDescription = "the number of milliseconds it took to evaluate all constraints for a template"
+
+	engineTypeName        = "engineType"
```
maxsmythe, line 39: This will collide with multi-engine:

https://github.com/maxsmythe/frameworks/blob/281f26c755af5da4c013e0b6eca3697807d655d3/constraint/pkg/client/drivers/rego/driver.go#L85-L88

A couple other changes also conflict. That being merged might be a blocker for this change.

constraint/pkg/client/drivers/local/driver.go
maxsmythe: In terms of accuracy and consistency. In general floats are harder to round-trip accurately:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

constraint/pkg/client/drivers/local/driver.go
maxsmythe: I see, how do users know how to interpret "key"... is it the name of the template? The constraint?

Also, "engine" sticks out a bit here, these metrics seem like they'll lose coherence if they get remixed or combined with other primitives.

constraint/pkg/client/drivers/local/driver.go
acpana: Interesting, good points.

One problem with using `ints` that I ran into before when gauging duration of a function is when integer division kicks in for values < 1. So evaluations that take less than a millisecond, in our case, can come out as `0`. In practice, I understand this may not be an issue but, theoretically, reporting `[0-1)` values as `0` could happen. 

Wdyt? If we can live with this edge case, I'm happy to use an int here

constraint/pkg/client/drivers/local/driver.go
acpana: > "key"

from some of our chats, I renamed `key` to `scope` but I think your question still stands!

> ... is it the name of the template? The constraint?

in the past we used `ConstraintKey` type as the type. To my mind, a consumer could type check in a `switch` and handle the key types as appropriate. However, using the template name (of type string) now I can see how a consumer won't be able to know what the `string` type stands for now.

Should I add another string to the `StatsEntry` like `ScopeType` that states what the Scope is actually? Something like:

```
{
"Scope": "K8sUniqueServiceSelector",
"ScopeType": "TemplateName",
"Stats": [...]
}
``` 

> Also, "engine" sticks out a bit here, these metrics seem like they'll lose coherence if they get remixed or combined with other primitives.

can you please offer an example here? I have a few thoughts but I'd like to understand scenarios where the `engineType` is problematic before sharing.

constraint/pkg/client/drivers/local/driver.go
acpana: These consts could live out in a different package. Was there something in particular you wanted to suggest changing?

constraint/pkg/client/drivers/local/driver.go
acpana: Per our offline chat, I've started to use nanoseconds to get around the issue above!

constraint/pkg/client/drivers/local/driver.go
acpana: we've added `scope` to denote the "what kind of" and `statsFor` to denote which specifically of the kind the stats refer to.

constraint/pkg/client/drivers/interface.go
```diff
@@ -40,7 +39,7 @@ type Driver interface {
 	// Returns results for each violated Constraint.
 	// Returns a trace if specified in query options or enabled at Driver creation.
 	// Returns an error if there was a problem executing the Query.
-	Query(ctx context.Context, target string, constraints []*unstructured.Unstructured, review interface{}, opts ...QueryOpt) ([]*types.Result, *string, error)
+	Query(ctx context.Context, target string, constraints []*unstructured.Unstructured, review interface{}, opts ...QueryOpt) *QueryResponse
```
acpana, line 42: pull out error from `QueryResponse` to stay go idiomatic

constraint/pkg/client/client.go
```diff
@@ -596,6 +591,15 @@ func (c *Client) Dump(ctx context.Context) (string, error) {
 	return c.driver.Dump(ctx)
 }
 
+func (c *Client) GetDescriptionForStat(statName string) string {
```
acpana, line 594: add in `source` as first arg and use it for the engine type casting; no switch; there will be a map from driver name --> driver obj; consider making a keyable `source` struct, something that works as a map key

constraint/pkg/client/client.go
```diff
@@ -571,23 +573,16 @@ func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.Qu
...
 	}
 
 	return &types.Response{
-		Trace:   trace,
-		Target:  target,
-		Results: results,
+		Trace:        qr.Trace,
+		Target:       target,
+		Results:      qr.Results,
+		StatsEntries: qr.StatsEntries,
```
acpana, line 585: move `StatsEntries` to the root and add a label of `target` in a subsequent coalition in `Review()`; have labels be namespace-d so client labels don't get clobbered

constraint/pkg/client/client.go
acpana: I will take another pass at using the `d.Name` from https://github.com/open-policy-agent/frameworks/pull/293 once that goes in but leaving as is for now.

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -12,6 +12,8 @@ import (
 	"time"
 
 	"github.com/open-policy-agent/frameworks/constraint/pkg/apis/constraints"
+	"github.com/open-policy-agent/frameworks/constraint/pkg/instrumentation"
```
maxsmythe, line 15: consolidate imports

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -31,6 +33,12 @@ import (
 const (
 	libRoot   = "data.lib"
 	violation = "violation"
+
+	templateRunTimeNanosName         = "templateRunTimeNanos"
```
maxsmythe, line 37: Nanos -> "NS"

https://en.wikipedia.org/wiki/Nanosecond

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -31,6 +33,12 @@ import (
...
+	templateRunTimeMillisDescription = "the number of nanoseconds it took to evaluate all constraints for a template"
```
maxsmythe, line 38: stale variable name

constraint/pkg/client/client.go
```diff
@@ -616,7 +618,7 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
 // Review makes sure the provided object satisfies all stored constraints.
 // On error, the responses return value will still be populated so that
 // partial results can be analyzed.
-func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.QueryOpt) (*types.Responses, error) {
+func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.Opt) (*types.Responses, error) {
```
maxsmythe, line 621: QueryOpt seems more descriptive, we should leave as-is

constraint/pkg/client/client.go
```diff
@@ -767,6 +786,36 @@ func (c *Client) Dump(ctx context.Context) (string, error) {
...
+		// only handle engine source for now
+		return instrumentation.UnknownDescription
+	}
+
+	// this is written in a general form
+	// but it only works for rego drivers for now.
+	for dName, d := range c.drivers {
+		if strings.EqualFold(dName, source.Value) {
+			if source.Value == instrumentation.RegoSource.Value {
+				// no other drivers implement the GetDescriptionForStat yet
```
maxsmythe, line 800: Add it to the Driver interface, updating the test driver as well.

There should be no need for type casting here.

constraint/pkg/client/client.go
```diff
@@ -767,6 +786,36 @@ func (c *Client) Dump(ctx context.Context) (string, error) {
...
 
+func (c *Client) GetDescriptionForStat(source instrumentation.Source, statName string) string {
+	if source.Type != instrumentation.EngineSourceType {
+		// only handle engine source for now
+		return instrumentation.UnknownDescription
+	}
+
+	// this is written in a general form
+	// but it only works for rego drivers for now.
+	for dName, d := range c.drivers {
```
maxsmythe, line 797: This should not require a for loop. Drivers are indexed by name, so there can be a direct lookup.

constraint/pkg/client/drivers/query_opts.go
```diff
@@ -1,16 +1,41 @@
...
+		}
+		cfg.EngineCfg.TracingEnabled = enabled
+	}
+}
+
+type DriverCfg struct {
+	StatsEnabled bool
+}
+
+// Stats(true) enables the driver to gather evaluation stats for a single
```
maxsmythe, line 31: s/gather/return

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -31,6 +33,12 @@ import (
...
+
+	constraintCountName        = "constraintCount"
+	constraintCountDescription = "the number of constraints that were evaluated at the same time for the given constraint kind"
```
maxsmythe, line 41: delete: "at the same time"

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -250,27 +254,36 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
 	}
 
 	d.mtx.RLock()
 	defer d.mtx.RUnlock()
 
+	allOptions := &drivers.Options{}
+	for _, opt := range opts {
+		opt(allOptions)
+	}
+	driverCfg := allOptions.DriverCfg
```
maxsmythe, line 267: Extra complexity created by splitting the config into two structs.

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -297,25 +310,58 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
+				&instrumentation.StatsEntry{
+					Scope:    instrumentation.TemplateScope,
+					StatsFor: kind,
+					Stats: []*instrumentation.Stat{
+						{
+							Name:  templateRunTimeNanosName,
+							Value: uint64(evalEndTime.Nanoseconds()),
+							Source: instrumentation.Source{
+								Type:  instrumentation.EngineSourceType,
+								Value: instrumentation.RegoEngineSource,
```
maxsmythe, line 329: For uniformity, the engine name should be taken from:

https://github.com/open-policy-agent/frameworks/blob/745f2b50cfcb746bb30997d65e851d1b6ee0b9e9/constraint/pkg/client/drivers/rego/schema/schema.go#L12

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -297,25 +310,58 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
+							Name:  "TracingEnabled",
+							Value: d.traceEnabled || (engineCfg != nil && engineCfg.TracingEnabled),
+						},
+						{
+							Name:  "PrintEnabled",
+							Value: d.printEnabled,
+						},
+						{
+							Name:  "EnableExternalDataClientAuth",
+							Value: d.enableExternalDataClientAuth,
```
maxsmythe, line 352: Let's omit this for now. It should be easy to infer by other means and should not impact performance.

constraint/pkg/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,54 @@
...
+const (
+	// scope constants.
+	TemplateScope = "template"
+
+	// description constants.
+	UnknownDescription = "unknown description"
+
+	// source constants.
+	EngineSourceType = "engine"
+	RegoEngineSource = "rego"
```
maxsmythe, line 14: This value should be defined by the engine, it should not live elsewhere in the code. Currently defined here:

https://github.com/open-policy-agent/frameworks/blob/745f2b50cfcb746bb30997d65e851d1b6ee0b9e9/constraint/pkg/client/drivers/rego/schema/schema.go#L12

constraint/pkg/client/e2e_test.go
```diff
@@ -834,24 +835,103 @@ func TestE2E_Tracing_Unmatched(t *testing.T) {
...
+			}
+
+			stats := rsps.StatsEntries
+			if stats == nil && tt.statsEnabled {
+				t.Fatal("got nil stats but stats enabled for Review")
+			} else if len(stats) != 0 && !tt.statsEnabled {
+				t.Fatal("got stats but stats disabled")
+			}
+
+			_, err = c.AddData(ctx, &obj.Object)
```
maxsmythe, line 882: Why is AddData() called? What is it testing?

constraint/pkg/client/e2e_test.go
```diff
@@ -834,24 +835,103 @@ func TestE2E_Tracing_Unmatched(t *testing.T) {
...
+				t.Fatal(err)
+			}
+		})
+	}
+}
+
+// TestE2E_Review_StatsEntries tests that we can get stats out of evaluated constraints.
+// In particular, this test makes sure we have stats for both violating constraint kind, name pairs
+// and non violating ones.
+func TestE2E_Review_StatsEntries(t *testing.T) {
```
maxsmythe, line 893: This test should iterate over a list of test cases. Each test case should have an inbound object and an expected stats object.

This will make what is being tested and the expected result more clear.

constraint/pkg/client/e2e_test.go
```diff
@@ -866,21 +946,69 @@ func TestE2E_Review_RegoEvaluationMeta(t *testing.T) {
...
-		stats, ok := result.EvaluationMeta.(rego.EvaluationMeta)
-		if !ok {
-			t.Fatalf("could not type convert to RegoEvaluationMeta")
-		}
+			for _, stat := range se.Stats {
+				// now check that descriptions are able to be fetched.
+				desc := c.GetDescriptionForStat(instrumentation.RegoSource, stat.Name)
+
+				switch stat.Name {
+				case "templateRunTimeNanos":
```
maxsmythe, line 960: Testing stat -> description should be a different test.

constraint/pkg/client/e2e_test.go
```diff
@@ -834,24 +835,103 @@ func TestE2E_Tracing_Unmatched(t *testing.T) {
...
+
+// TestE2E_Review_StatsEntries tests that we can get stats out of evaluated constraints.
+// In particular, this test makes sure we have stats for both violating constraint kind, name pairs
+// and non violating ones.
+func TestE2E_Review_StatsEntries(t *testing.T) {
 	ctx := context.Background()
-	c := clienttest.New(t)
-	ct := clienttest.TemplateCheckData()
-	_, err := c.AddTemplate(ctx, ct)
+	d, err := rego.New(rego.GatherStats())
```
maxsmythe, line 895: We should move away from using Rego to using the fake driver in tests. This will reduce randomness and allow us to do things like inject specific stats responses, error responses, etc.

maxsmythe (changes requested):

constraint/pkg/client/e2e_test.go
acpana: this shouldn't have been there, thanks for the spot!

constraint/pkg/client/drivers/query_opts.go
```diff
@@ -1,16 +1,41 @@
 package drivers
 
-type QueryCfg struct {
+type Options struct {
+	DriverCfg *DriverCfg
```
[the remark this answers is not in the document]
acpana: Consolidating reply re splitting the cfgs out here ... Initially, I thought that the EngineCfg was passed as was to the rego builder but that wasn't the case! Thanks for the call out as I too appreciate the simpler approach

constraint/pkg/client/e2e_test.go
acpana: I'll be adding a tt style closer to the driver, in the `driver_unit_test.go`. I think this end to end is still valuable but has grown in scope and will trim it down to just show that stats are received in a follow up commit.

constraint/pkg/client/e2e_test.go
acpana: I'm with you with using the `fake` driver for more int tests. In this case, what would be the best way to add in an automated int test for the `Rego` driver, in particular? 

I want to make sure that we get stats out of that one and no future changes to its internal cause a regression to this behavior.

maxsmythe: Can we write those tests next to the Rego driver, calling it directly?

acpana: For using the `fake` driver more in the e2e tests I've added: https://github.com/open-policy-agent/frameworks/issues/317 . In the case of this PR, I moved the `rego` tests in the driver package and kept the `DriverCfg` test to make sure that stats are returned when turned on.

constraint/pkg/client/e2e_test.go
acpana: globbed into the `cmp` usage

maxsmythe (approved):
LGTM, great job!

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -297,25 +306,54 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
+							Value: len(kindConstraints),
+							Source: instrumentation.Source{
+								Type:  instrumentation.EngineSourceType,
+								Value: schema.Name,
+							},
+						},
+					},
+					Labels: []*instrumentation.Label{
+						{
+							Name:  "TracingEnabled",
```
sozercan, line 339: do we want to make `TracingEnabled` and `PrintEnabled` into consts since they are repeated in a few other places?

sozercan (approved):
minor comments, otherwise LGTM

constraint/pkg/client/client.go
```diff
@@ -767,6 +785,25 @@ func (c *Client) Dump(ctx context.Context) (string, error) {
 	return dumpBuilder.String(), nil
 }
 
+func (c *Client) GetDescriptionForStat(source instrumentation.Source, statName string) string {
```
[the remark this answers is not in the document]
acpana: added an e2e test for the different sources and valid descriptions

constraint/pkg/client/client.go
```diff
@@ -767,6 +785,25 @@ func (c *Client) Dump(ctx context.Context) (string, error) {
...
+		return instrumentation.UnknownDescription
+	}
+
+	driver, ok := c.drivers[source.Value]
+	if !ok {
+		return instrumentation.UnknownDescription
+	}
+
+	desc, err := driver.GetDescriptionForStat(statName)
+	if err != nil {
```
ritazh, line 800: do we want to bubble this error up?

ritazh (approved):
minor comment
LGTM

constraint/pkg/client/client.go
acpana: I could be persuaded to bubble the error up. I know that Max made a comment to bubble up the error from the `driver.GetDescriptionForStat`. 

For the client UX, I thought it was better for the emission (logging) to not have to handle the error and always get in hand a string, the description requested or an "unknown description" string.

maxsmythe: IMO we should go all-in one direction or the other (either bubble up the error all the way or don't raise an error). 

For now, I'm okay with leaving this as-is, we can revamp as we learn more about how this is used.

PR #293, opened 2023-02-16
Add multi-engine support

maxsmythe:
This PR allows us to have multiple engines registered with a constraint framework client, which is a prerequisite for integrating with the CEL KEP:

https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/3488-cel-admission-control


constraint/pkg/client/drivers/remote/httpclient.go
```diff
@@ -1,289 +0,0 @@
-// Copyright 2017 The OPA Authors.  All rights reserved.
-// Use of this source code is governed by an Apache2
-// license that can be found in the LICENSE file.
-
-package remote
```
acpana, line 5: so are we writing off the `remote` driver here? or does it not make sense anymore? is it an unsupported case now?

acpana (commented):
taking a first pass; overall looks solid! excited to see this in

constraint/pkg/client/drivers/dummy/dummy.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (d *Driver) Query(ctx context.Context, target string, constraints []*unstructured.Unstructured, review interface{}, opts ...drivers.QueryOpt) ([]*types.Result, *string, error) {
+	results := []*types.Result{}
+	for i := range constraints {
+		constraint := constraints[i]
+		result := &types.Result{
+			Msg:        fmt.Sprintf("rejected by driver %s: %s", d.name, d.code[strings.ToLower(constraint.GetObjectKind().GroupVersionKind().Kind)]),
+			Constraint: constraint,
+			// TODO: the engine should not determine the enforcement action -- that does not work with CEL KEP
```
sozercan, line 207: highlighting TODO

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -0,0 +1,198 @@
+package k8scel
```
sozercan, line 1: you mentioned this is pending controller-runtime update?

constraint/pkg/client/client.go
```diff
@@ -135,29 +166,72 @@ func (c *Client) AddTemplate(ctx context.Context, templ *templates.ConstraintTem
...
 
-	if err := c.driver.AddTemplate(ctx, templ); err != nil {
+	newDriverN := c.driverForTemplate(templ)
+
+	driver, ok := c.drivers[newDriverN]
+	if !ok {
+		return resp, fmt.Errorf("%w: available drivers: %v, wanted %q", clienterrors.ErrNoDriver, c.driverPriority, c.driverForTemplate(templ))
+	}
+
+	// TODO: because different targets may have different code sets,
```
sozercan, line 176: highligting TODO. you mentioned this is used for shared cache, since this is out of scope for CEL at this time, can we create an issue for this?

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
```diff
@@ -65,6 +65,26 @@ spec:
...
+                        properties:
+                          engine:
+                            description: 'The engine used to evaluate the code. Example:
+                              "Rego". Required.'
+                            type: string
+                          source:
+                            description: The source code for the template. Required.
+                            x-kubernetes-preserve-unknown-fields: true
+                        required:
+                        - engine
```
ritazh, line 82: Should `source` be required too? It says required in the description.

constraint/pkg/client/client.go
```diff
@@ -135,29 +166,72 @@ func (c *Client) AddTemplate(ctx context.Context, templ *templates.ConstraintTem
 		return resp, err
 	}
 
-	if err := c.driver.AddTemplate(ctx, templ); err != nil {
+	newDriverN := c.driverForTemplate(templ)
+
+	driver, ok := c.drivers[newDriverN]
+	if !ok {
+		return resp, fmt.Errorf("%w: available drivers: %v, wanted %q", clienterrors.ErrNoDriver, c.driverPriority, c.driverForTemplate(templ))
+	}
```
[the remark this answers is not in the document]
maxsmythe: Then we wouldn't be able to use the return value of driverForTemplate() to generate an error string below.

We also would not be able to run a comparison to see if the driver changed (strings can always be trivially compared, structs may not be).

constraint/pkg/client/client.go
maxsmythe: This is a different TODO. This one is about supporting multi-target templates.

I probably wouldn't create a bug for this, since we are not actively working on multi-target templates ATM, and this would be part of that larger body of work. Having the TODO in the code is useful to avoid missing this nuance if/when we start working on it (issues are more likely to be missed).

constraint/pkg/client/client.go
```diff
@@ -135,29 +166,72 @@ func (c *Client) AddTemplate(ctx context.Context, templ *templates.ConstraintTem
...
+	// Remove old drivers last so that templates can be enforced
+	// despite a botched update
+	for oldDriverN := range cacheEntry.activeDrivers {
+		if oldDriverN == newDriverN {
+			continue
+		}
+		oldDriver, ok := c.drivers[oldDriverN]
+		if !ok {
+			return resp, fmt.Errorf("%w: while changing drivers", clienterrors.ErrNoDriver)
+		}
```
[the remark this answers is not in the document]
maxsmythe: Failing early is simpler, since there would only be one error to return. Continuing iteration means risking adding more errors.

Because the recovery modes here are:

* keep retrying and hope the failure goes away
* destroy the client and rebuild from scratch

there is little functional difference between the two approaches.

You could argue that it's then possible for every registered driver to have stale state in very specific circumstances if neither of the recovery modes are satisfied, but in practice the maximum # of drivers are hardcoded, and if this mode is common, there is likely a larger issue.

constraint/pkg/client/client.go
```diff
@@ -574,14 +687,41 @@ func (c *Client) review(ctx context.Context, target string, constraints []*unstr
...
+		template, ok := c.templates[strings.ToLower(constraint.GetObjectKind().GroupVersionKind().Kind)]
+		if !ok {
+			return nil, fmt.Errorf("%w: while loading driver for constraint %s", ErrMissingConstraintTemplate, constraint.GetName())
+		}
+		driver := c.driverForTemplate(template.template)
+		if driver == "" {
+			return nil, fmt.Errorf("%w: while loading driver for constraint %s", clienterrors.ErrNoDriver, constraint.GetName())
+		}
+		driverToConstraints[driver] = append(driverToConstraints[driver], constraint)
+	}
```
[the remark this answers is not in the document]
maxsmythe: Is there any benefit to doing that? That seems like a purely lateral move.

In a future where we have multi-target templates, it's possible that each target will have different sets of source code, which means the results of `driverForTemplate()` would change depending on which target we are looking at. This means the mapping from constraint -> driver has to happen after we know what template we are looking at.

constraint/pkg/client/client.go
```diff
@@ -574,14 +687,41 @@ func (c *Client) review(ctx context.Context, target string, constraints []*unstr
...
+
+	for driverName, driver := range c.drivers {
+		if len(driverToConstraints[driverName]) == 0 {
+			continue
+		}
+		driverResults, trace, err := driver.Query(ctx, target, driverToConstraints[driverName], review, opts...)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
maxsmythe: In this case it probably makes sense to be able to return as much valid data as possible.

constraint/pkg/client/client_opts.go
```diff
@@ -54,7 +54,11 @@ func validateTargetNames(ts []handler.TargetHandler) []string {
 // Driver defines the Rego execution environment.
 func Driver(d drivers.Driver) Opt {
 	return func(client *Client) error {
-		client.driver = d
+		if d.Name() == "" {
+			return ErrNoDriverName
+		}
+		client.drivers[d.Name()] = d
+		client.driverPriority[d.Name()] = len(client.drivers)
```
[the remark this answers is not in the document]
maxsmythe: Not really.

There is the argument that it's harder to keep the state of two different maps in sync, except this function is the only place where these maps get altered.

This is counterbalanced against the fact that it'd be slightly harder to write something like driverForTemplate(), since you wouldn't be able to rely on conventions like a missing priority entry == 0, plus you'd need an additional struct.

In the end it's a wash.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: You could call driver.Name(), but I really don't see any benefit to doing it the way you suggest.

constraint/pkg/client/drivers/dummy/dummy.go
```diff
@@ -0,0 +1,217 @@
+package dummy
```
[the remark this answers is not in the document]
maxsmythe: May I ask why?

constraint/pkg/client/drivers/dummy/dummy.go
maxsmythe: LMK if you want to see this done in this PR, I figured it'd be better as a follow-on

constraint/pkg/client/drivers/interface.go
```diff
@@ -11,6 +11,10 @@ import (
 
 // A Driver implements Rego query execution of Templates and Constraints.
 type Driver interface {
+	// Name returns the name of the driver, used to uniquely identify a driver
+	// and in errors returned to the user.
+	Name() string
```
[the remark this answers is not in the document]
maxsmythe: No, we need Name() as a unique string identifier for a driver so that:

* Constraint templates can cite the driver name in their source code
* We can dereference that name into a specific driver instance
* We have a reliable comparator to see if a driver has changed.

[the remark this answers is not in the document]
maxsmythe: Because Driver is purely for internal consumption, if we were going to require extra data in order for driver to be useful, that is a de-facto breaking change.

If we wanted to add some kind of optional feature, it should not be in the interface, because then it would be required (typecasting can be used to convert to a more feature-rich interface if we want to support optional methods).

constraint/pkg/client/drivers/k8scel/driver.go
maxsmythe: Correct. We can't use the K8s CEL code without the change that makes the result objects public (which is bleeding edge), and we can't use bleeding-edge k8s code until controller-runtime supports it.

constraint/pkg/client/drivers/rego/driver.go
```diff
@@ -73,14 +74,19 @@ type Driver struct {
 	clientCertWatcher *certwatcher.CertWatcher
 }
 
-// RegoEvaluationMeta has rego specific metadata from evaluation.
-type RegoEvaluationMeta struct {
+// EvaluationMeta has rego specific metadata from evaluation.
+type EvaluationMeta struct {
```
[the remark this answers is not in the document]
maxsmythe: That would make the linter angry

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: Also, if this winds up being the wrong choice, it can be changed.

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
maxsmythe: good catch, apparently "omitempty" overrides the controller-gen flag, fixed.

constraint/pkg/client/drivers/remote/httpclient.go
maxsmythe: It has been dead code for a long time. Removing it.

constraint/pkg/client/client_opts.go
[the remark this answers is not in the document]
ritazh: If d.Name() already exists as part of client.drivers, could this alter the driverPriority since the total len could be different?

constraint/pkg/client/client.go
```diff
@@ -93,7 +123,9 @@ func (c *Client) AddTemplate(ctx context.Context, templ *templates.ConstraintTem
 		}
 	}
 
-	if cachedCpy != nil && cachedCpy.SemanticEqual(templ) {
+	// if there is more than one active driver for the template, there is some cleanup to do
+	// from a botched driver swap.
+	if cachedCpy != nil && cachedCpy.SemanticEqual(templ) && len(cached.activeDrivers) == 1 {
```
ritazh, line 128: why is this only true when there is only 1 activeDrivers?

constraint/pkg/client/client.go
```diff
@@ -135,29 +167,72 @@ func (c *Client) AddTemplate(ctx context.Context, templ *templates.ConstraintTem
...
 	// a panic.
-	if template == nil {
-		template = &templateClient{
-			constraints: make(map[string]*constraintClient),
+	if cacheEntry == nil {
+		cacheEntry = newTemplateClient()
+		c.templates[templateName] = cacheEntry
+	}
+
+	cacheEntry.activeDrivers[newDriverN] = true
```
ritazh, line 196: is activeDrivers thread safe? does it need any locks?

constraint/pkg/client/client.go
```diff
@@ -189,22 +264,33 @@ func (c *Client) RemoveTemplate(ctx context.Context, templ *templates.Constraint
...
+	// to ensure cleanup in case of a botched update
+	for driverN := range cached.activeDrivers {
+		driver, ok := c.drivers[driverN]
+		if !ok {
+			return resp, fmt.Errorf("%w: could not clean up %q", clienterrors.ErrNoDriver, driverN)
+		}
+
+		err := driver.RemoveTemplate(ctx, template)
+		if err != nil {
+			return resp, err
```
ritazh, line 286: return early here prevents it from removing from other drivers. is this intentional?

constraint/pkg/client/client.go
```diff
@@ -431,14 +533,20 @@ func (c *Client) AddData(ctx context.Context, data interface{}) (*types.Response
...
 		}
 
-		err = c.driver.AddData(ctx, name, key, processedDataCpy)
-		if err != nil {
-			errMap[name] = err
+		// To avoid maintaining duplicate caches, only Rego should get its own
+		// storage. We should work to remove the need for this special case
+		// by building a global storage object. Right now Rego needs its own
+		// cache to cache constraints.
+		if _, ok := c.drivers[regoSchema.Name]; ok {
```
ritazh, line 540: what happens if !ok is returned?

constraint/pkg/client/client.go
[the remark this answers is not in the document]
acpana: >  to generate an error string below.

we could wrap the err returned by `driverForTemplate` to add more contextual information as needed? 

I think `driverForTemplate` is misleading since we actually return the language/ name of the driver out of that func and not the `driver` struct/ object itself. Maybe `diverNameForTemplate` otherwise?

whatever you choose here is fine w me

constraint/pkg/client/client_opts.go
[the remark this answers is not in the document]
maxsmythe: Good call, I should have checked for uniqueness.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: I think it is not worth the time to bikeshed on the naming of purely internal functions.

constraint/pkg/client/drivers/dummy/dummy.go
maxsmythe: ping @sozercan

constraint/pkg/client/client.go
maxsmythe: If the contents haven't change and there is only one active driver, there is no work to do and we can return early.

If the contents have changed, then there is work to do to update the template code.

If the contents have not change, but there is more than one active driver, then there is cleanup to do (only one driver should be active at a time).

If the contents have not changed, and there are zero active drivers... this state shouldn't happen, but we should add the template code to a driver so we have an active driver.

constraint/pkg/client/client.go
maxsmythe: It would not be thread safe, but I believe `cacheEntry` is gated by the `c.mtx` lock above?

constraint/pkg/client/client.go
maxsmythe: Yep, the idea is that a failed removal needs to be met with a retry (as in a controller). If we cannot fully remove a template from a driver and cannot recover from that state, then the system is in an unknown state and the program should die (e.g. we have stale Rego code in OPA, who knows what impact it may have)

constraint/pkg/client/client.go
maxsmythe: This code is just saying "only call add/remove data for the Rego driver".

Having multiple drivers maintain separate caches would lead to excess RAM usage.

Ultimately, what we need is a cache that can be shared across drivers, but that is quite a bit of work, and not necessary for the CEL KEP (which has no notion of referential constraints).

constraint/pkg/client/drivers/dummy/dummy.go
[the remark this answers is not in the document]
maxsmythe: Still not sure of the reasoning, but to avoid bikeshedding, renamed to "fake"

constraint/pkg/client/drivers/dummy/dummy.go
sozercan: i am okay to defer, just want to capture this and attach an issue so we don't forget

maxsmythe: Added #298

constraint/pkg/client/client.go
maxsmythe: LMK if you'd prefer the issue though

acpana (approved):
LGTM! Thanks for addressing the comments I made 🥇

sozercan (approved):
Thank you! LGTM

sozercan:
@maxsmythe do you mind if we merge https://github.com/open-policy-agent/frameworks/pull/300 first so we can cut a release branch and tag for v3.12?

constraint/pkg/client/client.go
ritazh: should we at least log it if it's not a rego driver?

ritazh:
Seems gatekeeper_test  is failing?

ritazh (approved):
LGTM
one small nit and test failure

maxsmythe:
gatekeeper test is failing b/c of the renaming of the "local" driver package to "rego", so is expected.

constraint/pkg/client/client.go
maxsmythe: My concern there would be log verbosity. For example: someone who uses Gatekeeper who watches pods would see a log line every time a pod was updated.

maxsmythe: OTOH, Gatekeeper would have the rego driver, so no need to log

ritazh: https://github.com/open-policy-agent/frameworks/pull/293/commits/19851c7e118f61a744af77636fed70807001cf4d LGTM. Can you pls resolve the file conflict?

maxsmythe: Thanks! Resolved.

PR #299, opened 2023-03-01
rework tracedump msg

acpana:
If a review doesn't match a constraint's match spec AND tracing is enabled, then we print `TRACE: TRACING DISABLED`. This can confuse consumers because they specifically turn on tracing. We should only print this message if there are results and the trace is nil. See non test changes for in code comments too.

### DIFF
--- a/constraint/pkg/types/validation.go
+++ b/constraint/pkg/types/validation.go
@@ -65,7 +65,15 @@ func (r *Response) TraceDump() string {
 	b := &strings.Builder{}
 	_, _ = fmt.Fprintf(b, "Target: %s\n", r.Target)
 	if r.Trace == nil {
-		_, _ = fmt.Fprintf(b, "Trace: TRACING DISABLED\n\n")
+		if r.Results != nil {
+			// only say "Trace: TRACING DISABLED" if there are results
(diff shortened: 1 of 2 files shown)


maxsmythe (approved):
LGTM

sozercan (approved):
LGTM

constraint/pkg/types/validation.go
```diff
@@ -65,7 +65,15 @@ func (r *Response) TraceDump() string {
...
 	if r.Trace == nil {
-		_, _ = fmt.Fprintf(b, "Trace: TRACING DISABLED\n\n")
+		if r.Results != nil {
+			// only say "Trace: TRACING DISABLED" if there are results
+			// otherwise, we risk to confuse consumers who see the msg
+			// and think that evaluation did not happen.
+
+			// Note that if there were NO violating results AND the trace
+			// was turned on, then r.Trace != nil.
+			_, _ = fmt.Fprintf(b, "Trace: TRACING DISABLED\n\n")
```
ritazh, line 75: nit: make this a const that you can also test in unit test?

acpana: added in!

PR #322, opened 2023-04-13
chore: add assert/require testing lib

acpana:
Also replace some usage in `e2e_test` for the import not to be auto removed.


maxsmythe:
I am against this change. There is no reason to change the unit test libraries here, as such this introduces change risk with no benefit.

If new testing is being written, and you want to use the assert/require libraries, I have no objection to that.

acpana:
> If new testing is being written, and you want to use the assert/require libraries, I have no objection to that.

I'd've liked to use this library when working on stats but decided against importing it to keep the diff small. I can see myself using this in future frameworks PRs. Rather than adding this lib as a standalone, it seems the suggestion here is to add it the next time I want to use it ?

maxsmythe:
Yep!

Issue #70, opened 2020-01-15
Properly handle OpenAPISpec Default values


briantkennedy:
If someone puts a default value in the OpenAPISpec for a ConstraintTemplate, this results in an error.  The ConstraintFramework should handle populating the default if present.

acpana:
I'm not sure this is still the case today, but no errors are returned when attempting to use `default`s w OpenAPIV3 schema for a `ConstraintTemplate`.

acpana:
I'm sure the ask is the same: make `default` params work but just in case I filed another issue: https://github.com/open-policy-agent/frameworks/issues/326

PR #325, opened 2023-04-22
fix: default params for open api

acpana:
Fixes https://github.com/open-policy-agent/frameworks/issues/326.
Fixes https://github.com/open-policy-agent/frameworks/issues/70.

Inject any defined default parameters from the ConstraintTemplate schema definition in a Constraint that does not define its parameters.

### DIFF
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -358,13 +358,18 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 		return resp, clienterrors.ErrNoDriver
 	}
 
-	changed, err := cached.AddConstraint(constraint)
+	constraintWDefaults, err := c.withDefaultParams(constraint, template)
+	if err != nil {
+		return resp, err
+	}
+
+	changed, err := cached.AddConstraint(constraintWDefaults)
 	if err != nil {
 		return resp, err
 	}
 
 	if changed {
-		err = driver.AddConstraint(ctx, constraint)
+		err = driver.AddConstraint(ctx, constraintWDefaults)
 		if err != nil {
 			return resp, err
 		}
@@ -375,6 +380,51 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 	}
 
 	return resp, nil
+}
+
+// withDefaultParams injects any default values defined in the OpenAPIV3Schema
+// to the constraint spec only if a parameter is not already defined in the constraint.
+//
+//	Returns a copy of the constraint with the defaults or an error in processing the schema and constraint.
+func (c *Client) withDefaultParams(constraint *unstructured.Unstructured, templ *templates.ConstraintTemplate) (*unstructured.Unstructured, error) {
+	if templ.Spec.CRD.Spec.Validation == nil || templ.Spec.CRD.Spec.Validation.LegacySchema != nil {
+		return constraint, nil
+	}
+
+	cpy := constraint.DeepCopy()
+	defaults := map[string]*apiextensions.JSON{}
+	schema := templ.Spec.CRD.Spec.Validation.OpenAPIV3Schema
+
+	for paramName := range schema.Properties {
+		paramProps := schema.Properties[paramName]
+
+		if paramProps.Default != nil {
+			defaults[paramName] = paramProps.Default
+		}
+	}
+
+	_, _, err := unstructured.NestedFieldNoCopy(cpy.Object, "spec", "parameters")
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", apiconstraints.ErrInvalidConstraint, err)
+	}
+
+	for paramName, defaultParam := range defaults {
+		// only look to default parameters that are not found
+		_, found, err := unstructured.NestedFieldNoCopy(cpy.Object, "spec", "parameters", paramName)
+		if err != nil {
+			return nil, fmt.Errorf("could not find parameter %s in CRD %w: %v", paramName, apiconstraints.ErrInvalidConstraint, err)
+		}
+
+		// by definition, if we looked up a paramName, it was a parameter
+		// that had a Default defined in the schema so we are right to default it.
+		if !found {
+			if err := unstructured.SetNestedField(cpy.Object, *defaultParam, "spec", "parameters", paramName); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return cpy, nil
 }
 
 // RemoveConstraint removes a constraint from OPA. On error, the responses
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -358,18 +358,18 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 		return resp, clienterrors.ErrNoDriver
 	}
 
-	constraintWDefaults, err := c.withDefaultParams(constraint, template)
+	constraintWithDefaults, err := c.withDefaultParams(constraint, template)
 	if err != nil {
 		return resp, err
 	}
 
-	changed, err := cached.AddConstraint(constraintWDefaults)
+	changed, err := cached.AddConstraint(constraintWithDefaults)
 	if err != nil {
 		return resp, err
 	}
 
 	if changed {
-		err = driver.AddConstraint(ctx, constraintWDefaults)
+		err = driver.AddConstraint(ctx, constraintWithDefaults)
 		if err != nil {
 			return resp, err
 		}
@@ -387,7 +387,7 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 //
 //	Returns a copy of the constraint with the defaults or an error in processing the schema and constraint.
 func (c *Client) withDefaultParams(constraint *unstructured.Unstructured, templ *templates.ConstraintTemplate) (*unstructured.Unstructured, error) {
-	if templ.Spec.CRD.Spec.Validation == nil || templ.Spec.CRD.Spec.Validation.LegacySchema != nil {
+	if templ.Spec.CRD.Spec.Validation == nil || templ.Spec.CRD.Spec.Validation.OpenAPIV3Schema == nil {
 		return constraint, nil
 	}
 
--- a/constraint/pkg/client/client.go
+++ b/constraint/pkg/client/client.go
@@ -323,8 +323,8 @@ func (c *Client) GetTemplate(templ *templates.ConstraintTemplate) (*templates.Co
 	return template.getTemplate(), nil
 }
 
-// getTemplateEntry returns the template entry for a given constraint.
-func (c *Client) getTemplateForKind(kind string) *templateClient {
+// getTemplateClientForKind returns the template entry for a given constraint.
+func (c *Client) getTemplateClientForKind(kind string) *templateClient {
 	name := strings.ToLower(kind)
 
 	return c.templates[name]
@@ -345,7 +345,7 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 	}
 
 	kind := constraint.GetKind()
-	cached := c.getTemplateForKind(kind)
+	cached := c.getTemplateClientForKind(kind)
 	if cached == nil {
 		templateName := strings.ToLower(kind)
 		return resp, templateNotFound(templateName)
@@ -358,7 +358,7 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 		return resp, clienterrors.ErrNoDriver
 	}
 
-	constraintWithDefaults, err := c.withDefaultParams(constraint, template)
+	constraintWithDefaults, err := cached.ApplyDefaultParams(constraint)
 	if err != nil {
 		return resp, err
 	}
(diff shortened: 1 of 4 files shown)


constraint/pkg/client/client.go
```diff
@@ -358,13 +358,18 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 		return resp, clienterrors.ErrNoDriver
 	}
 
-	changed, err := cached.AddConstraint(constraint)
+	constraintWDefaults, err := c.withDefaultParams(constraint, template)
```
maxsmythe, line 361: let's just call it `constraintWithDefaults`, the "W" is more confusing than illuminating otherwise

constraint/pkg/client/client.go
```diff
@@ -377,6 +382,51 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 	return resp, nil
 }
 
+// withDefaultParams injects any default values defined in the OpenAPIV3Schema
+// to the constraint spec only if a parameter is not already defined in the constraint.
+//
+//	Returns a copy of the constraint with the defaults or an error in processing the schema and constraint.
+func (c *Client) withDefaultParams(constraint *unstructured.Unstructured, templ *templates.ConstraintTemplate) (*unstructured.Unstructured, error) {
+	if templ.Spec.CRD.Spec.Validation == nil || templ.Spec.CRD.Spec.Validation.LegacySchema != nil {
```
maxsmythe, line 390: Why are we bypassing if legacySchema is defined?

constraint/pkg/client/client.go
```diff
@@ -358,13 +358,18 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
 		return resp, clienterrors.ErrNoDriver
 	}
 
-	changed, err := cached.AddConstraint(constraint)
+	constraintWDefaults, err := c.withDefaultParams(constraint, template)
+	if err != nil {
+		return resp, err
+	}
+
+	changed, err := cached.AddConstraint(constraintWDefaults)
```
maxsmythe, line 366: IIRC cached.AddConstraint also deepcopies... is there a way to coherently eliminate redundant deepcopies?

constraint/pkg/client/client.go
```diff
@@ -377,6 +382,51 @@ func (c *Client) AddConstraint(ctx context.Context, constraint *unstructured.Uns
...
+			defaults[paramName] = paramProps.Default
+		}
+	}
+
+	_, _, err := unstructured.NestedFieldNoCopy(cpy.Object, "spec", "parameters")
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", apiconstraints.ErrInvalidConstraint, err)
+	}
+
+	for paramName, defaultParam := range defaults {
```
maxsmythe, line 411: Does K8s have this logic already? Can we call that logic?

maxsmythe (changes requested):

constraint/pkg/client/client.go
acpana: At first glance, I did not find something to use in `apiextensions` but maybe I wasn't looking in the right place? I've looked at `SetDefaults_x` and how we have an existing wired up way to add some defaults in CF but I wasn't sure if they could be used for the defaults that I care about here.

constraint/pkg/client/client.go
acpana: It does! 

https://github.com/open-policy-agent/frameworks/blob/d82cbe189211c80b37f0d827ca813b39c9da05ce/constraint/pkg/client/template_client.go#L101-L110

the `cpy` above seems to actually be stored (if changed/ different).

the `cpy` made from the `withDefaultParams` should get garbage collected (eventually) when this func exists as it's not stored in its entirety in `driver.AddConstraint()` so I didn't think we'd take a performance hit. Is there a particular concern here?

---

That said, more than happy to make some changes. We could eliminate the deep cpy from template_client.AddConstraint?

constraint/pkg/client/client.go
acpana: Good catch! This should be `OpenAPIV3Schema`.  I think early when I looked at this I was unsure what the relationship b/w `LegacySchema` and the actual openapiv3 spec was and figured they may be mutually exclusive! I know that's not the case now.

This line should actually make sure we have `Validation` and `OpenAPIV3Schema` available so we don't panic

acpana:
@maxsmythe thanks for reviewing 💯 ; left some questions and, otherwise, incorporated feedback. PTAL when you have some time 🙏🏼

davis-haba:
nice! lgtm

acpana:
@davis-haba @maxsmythe As suggested, I changed the guts to use `defaulting.Default` to set the default parameters if they are defined in the CRD spec. PTAL 🙏🏼 

Also, FYI it seems this way of doing it adds an empty param map which breaks some gator `TestTest` cases. 

https://github.com/open-policy-agent/frameworks/actions/runs/4888449293/jobs/8726107599?pr=325#step:6:846

IMO, this is fine -- maybe the tests shouldn't be so brittle and am willing to go fix them as part of the CF upgrade in g8r. What do y'all think?

constraint/pkg/client/template_client.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/open-policy-agent/frameworks/constraint/pkg/handler"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
```
maxsmythe, line 14: nit: no newline

constraint/pkg/client/template_client.go
```diff
@@ -60,6 +63,44 @@ func (e *templateClient) ValidateConstraint(constraint *unstructured.Unstructure
 	return crds.ValidateCR(constraint, e.crd)
 }
 
+// ApplyDefaultParams will apply any default parameters defined in the CRD of the constraint's
+// corresponding template.
+// Assumes ValidateConstraint() called is called so the constraint is a valid CRD.
+func (e *templateClient) ApplyDefaultParams(constraint *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	cpy := constraint.DeepCopy()
```
maxsmythe, line 70: We shouldn't deepcopy here, just modify the input value in-place. Let callers decide if deepcopy is necessary or if they'd prefer to avoid the performance penalty.

constraint/pkg/client/template_client.go
```diff
@@ -60,6 +63,44 @@ func (e *templateClient) ValidateConstraint(constraint *unstructured.Unstructure
...
+
+	params, found, err := unstructured.NestedFieldNoCopy(cpy.Object, "spec", "parameters")
+	if err != nil {
+		return nil, err
+	}
+	if !found {
+		params = make(map[string]interface{})
+	}
+
+	// sequentially get the structural schema for the spec field
```
maxsmythe, line 80: Any reason why we shouldn't apply all defaults specified by the CRD?

acpana: ( continuing our offline chat here)

> Any reason why we shouldn't apply all defaults specified by the CRD?

Not a particular reason but it seems that the other default we care about (legacy) works as expected so I was trying to be topical in only defaulting the OpenAPIV3 params.  

Without converting `e` or `e.crd` to a `apiextensions.JSONSchemaProps`, best we can do is :

```
	apiStructural, err := schema.NewStructural(e.crd.Spec.Validation.OpenAPIV3Schema)
	defaulting.Default(spec, apiStructural)

	if err := unstructured.SetNestedField(constraint.Object, spec, "spec"); err != nil {
		return nil, err
	}
```

Which, as discussed, doesn't work as one would think since it doesn't see a Default on the top level and bails early. 

---

Another take on this would be to look at how we default when we create the CRD:

https://github.com/open-policy-agent/frameworks/blob/3f237e2710faee505189afd605cfe75ed89dd446/constraint/pkg/client/crds/crds.go#L82-L85

I can look to explore that route but it may require some number of conversions IIUC, which adds more risk to this func. Also, I think it's worth having an "external" (as in external to the creation) func do this since `AddConstraint` can contain objects that are not constraints (which get weeded out on validation).

What do you think?

maxsmythe: `e.crd.Spec.Validation.OpenAPIV3Schema` describes the schema for the object root (not starting with spec).

To conform with the behavior of the K8s API server, let's apply the default to the root of the object. Doing this will make gator and `kubectl apply`-ing to a cluster behave identically (barring version skew WRT k8s defaulting code).

> Which, as discussed, doesn't work as one would think since it doesn't see a Default on the top level and bails early.

It works as K8s defaults work. Since we are piggybacking on that standard, following it is least surprising.

> Another take on this would be to look at how we default when we create the CRD

Not sure what you're getting at with this section. In any case, we shouldn't look to do too much in this PR... move slow, get feedback, iterate. There is more to lose by doing something that is broken than there is to gain by doing something quickly.

> Also, I think it's worth having an "external" (as in external to the creation) func

Given I didn't follow what you were saying, I may be missing something here, but we should not create an externally-accessible constraint defaulting function. I'm not sure what the use for this would be, and could be added at any time if it ever became necessary.

acpana: > To conform with the behavior of the K8s API server, let's apply the default to the root of the object.

Okay I can buy that, in general, but, in this case, only `default`-ing to the root object will have the practical effect on not actually defaulting the parameters. So the UX is still, practically, broken. We can document the this papercut but I think it may be worth paying the price here to do the right thing.

>  Doing this will make gator and kubectl apply-ing to a cluster behave identically

This may be an aside but AFAICT, `kubectl apply` today won't default the parameters either (or maybe i'm holding it wrong).

constraint/pkg/client/template_client.go
```diff
@@ -60,6 +62,42 @@ func (e *templateClient) ValidateConstraint(constraint *unstructured.Unstructure
 	return crds.ValidateCR(constraint, e.crd)
 }
 
+// ApplyDefaultParams will apply any default parameters defined in the CRD of the constraint's
+// corresponding template.
+// Assumes ValidateConstraint() called is called so the constraint is a valid CRD.
```
maxsmythe, line 67: "called is called"

constraint/pkg/client/template_client.go
maxsmythe: Per offline conversation, we're going to apply defaults at the root of the constraint object.

maxsmythe (changes requested):

constraint/pkg/client/template_client.go
```diff
@@ -60,6 +62,19 @@ func (e *templateClient) ValidateConstraint(constraint *unstructured.Unstructure
...
```
acpana, line 67:
```suggestion
// Assumes ValidateConstraint() is called so the constraint is a valid CRD.
```

constraint/pkg/client/client_test.go
```diff
@@ -1039,6 +1041,261 @@ func TestClient_AddConstraint(t *testing.T) {
...
+		// like for a constraint that actually defines a parameter.
+		constraintSpec map[string]interface{}
+
+		wantSpec               map[string]interface{}
+		wantHandled            map[string]bool
+		wantAddConstraintError error
+		wantGetConstraintError error
+
+		// driver is pulled out to test actual driver implementations.
+		// defulat is fake.
```
maxsmythe, line 1078: "defulat" -> "default"

constraint/pkg/client/client_test.go
```diff
@@ -1039,6 +1041,261 @@ func TestClient_AddConstraint(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			wantSpec: map[string]interface{}{
+				"parameters": map[string]interface{}{
+					"foo": defaults,
+				},
+			},
+			driver: regoDriver,
```
maxsmythe, line 1110: Why do we need to test this with anything other than the fake driver?

For ease of writing tests, we should be moving away from non-trivial drivers. This is especially true for this test, since the driver is not part of the relevant functionality.

constraint/pkg/client/client_test.go
```diff
@@ -1039,6 +1041,261 @@ func TestClient_AddConstraint(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			tc := tc
+			// given a template and constraint for it
+			template := template.DeepCopy()
```
maxsmythe, line 1229: let's not rely on shadowing. Let's be explicit, call the right side something like "baseTemplate", same for "constraint" below.

constraint/pkg/client/client_test.go
```diff
@@ -1039,6 +1041,261 @@ func TestClient_AddConstraint(t *testing.T) {
...
+		wantSpec               map[string]interface{}
+		wantHandled            map[string]bool
+		wantAddConstraintError error
+		wantGetConstraintError error
+
+		// driver is pulled out to test actual driver implementations.
+		// defulat is fake.
+		driver drivers.Driver
+	}{
+		// this no op test case is tested implicitly in all
```
maxsmythe, line 1081: no need for per-test comments.

maxsmythe (approved):
LGTM after nits

constraint/pkg/client/client_test.go
acpana: I find that comments on tests help me understand why the test is there to begin with. I'm curious if others feel the same. If you feel strongly against adding these, I can take the comments out.

constraint/pkg/client/client_test.go
acpana: I think it's worth having this functional test to prove that constraints with defaulted params are well handled in the rego driver.

All other cases are using the fake driver to test the flows.

constraint/pkg/client/client_test.go
maxsmythe: Strongly against. The name and parameters are descriptive enough, with the comments mostly duplicating the code. Comments that are tightly coupled to code also tend to go stale.

constraint/pkg/client/client_test.go
maxsmythe: We should use the fake driver only. This is not an integration test and furthermore the driver is never actually invoked. Finally, we should not be setting up an expectation that future drivers also need tests here (e.g. K8s admission driver).

maxsmythe (changes requested):
Removing LGTM, since it was dependent on nit fixes

maxsmythe (changes requested):
Removing LGTM, since it was dependent on nit fixes

maxsmythe:
Weird, GitHub double posted, sorry about that.

constraint/pkg/client/client_test.go
acpana: test case comments removed.

constraint/pkg/client/client_test.go
acpana: sounds good, i removed the rego driver from this test func; using fake driver only.

maxsmythe (approved):
LGTM, thanks!

acpana:
@sozercan @ritazh PTAL when you have a chance 🙏🏼 lmk if you have any questions

JaydipGabani (approved):
LGTM, nice work!

sozercan (approved):
LGTM

PR #334, opened 2023-06-08
feat: implements external data response cache

nilekhc:


constraint/pkg/client/drivers/rego/builtin.go
```diff
@@ -25,12 +31,65 @@ func externalDataBuiltin(d *Driver) func(bctx rego.BuiltinContext, regorequest *
...
+					)
+				}
+				prepareResponse.Items = append(prepareResponse.Items, externaldataResponse.Response.Items...)
+			}
+
+			providerResponseStatusCode = statusCode
+			prepareResponse.SystemError = externaldataResponse.Response.SystemError
+		}
+
+		prepareResponse.Idempotent = true
```
maxsmythe, line 85: Why are we setting idempotent to true?

constraint/pkg/externaldata/cache.go
```diff
@@ -16,6 +18,74 @@ type ProviderCache struct {
...
+
+func (c *ProviderResponseCache) Remove(key CacheKey) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
+	delete(c.cache, key)
+}
+
+func (c *ProviderResponseCache) invalidateProviderResponseCache(ctx context.Context, ttl time.Duration) {
+	for {
```
maxsmythe, line 74: Is there any throttling of this for-loop (a sleep, a timer tic, etc.)?

constraint/pkg/client/drivers/rego/builtin.go
nilekhc: Don't we need it to be `true` for mutation?

constraint/pkg/externaldata/cache.go
nilekhc: Updated code

constraint/pkg/client/drivers/rego/builtin.go
maxsmythe: Right, but it's an attestation from the external data source that it is idempotent. Hardcoding that it is defeats the purpose of the check.

nilekhc: Got it. Removed.

maxsmythe: Is the cache being used for mutation as well? We probably want to preserve the response from the provider somehow, if so.

nilekhc: @maxsmythe I am now preserving Idempotent from provider response.

constraint/pkg/client/drivers/rego/builtin.go
```diff
@@ -25,12 +31,78 @@ func externalDataBuiltin(d *Driver) func(bctx rego.BuiltinContext, regorequest *
...
+					prepareResponse.Items, externaldata.Item{
+						Key:   k,
+						Value: cachedResponse.Value,
+						Error: cachedResponse.Error,
+					},
+				)
+
+				// we are taking conservative approach here, if any of the cached response is not idempotent
+				// we will mark the whole response as not idempotent
+				if !cachedResponse.Idempotent && prepareResponse.Idempotent {
```
maxsmythe, line 61: no need for the second part of the && clause.

constraint/pkg/client/drivers/rego/builtin.go
```diff
@@ -25,12 +31,78 @@ func externalDataBuiltin(d *Driver) func(bctx rego.BuiltinContext, regorequest *
...
+						Received:   time.Now().Unix(),
+						Value:      item.Value,
+						Error:      item.Error,
+						Idempotent: externaldataResponse.Response.Idempotent,
+					},
+				)
+
+				// we are taking conservative approach here, if any of the response is not idempotent
+				// we will mark the whole response as not idempotent
+				if !externaldataResponse.Response.Idempotent && prepareResponse.Idempotent {
```
maxsmythe, line 89: This should be outside of the loop, since it is only set once in the response object. Also, no need for the second part of the if statement.

maxsmythe (approved):
LGTM after nits

constraint/pkg/client/drivers/rego/driver_unit_test.go
```diff
@@ -743,6 +744,7 @@ func TestDriver_ExternalData(t *testing.T) {
 
 			d, err := New(
 				AddExternalDataProviderCache(externaldata.NewCache()),
+				AddExternalDataProviderResponseCache(externaldata.NewProviderResponseCache(context.Background(), 1*time.Minute)),
```
sozercan, line 747: would we want this to be configurable?

nilekhc: TTL? It is configurable - https://github.com/open-policy-agent/gatekeeper/pull/2823/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R112

sozercan: I missed that this is for tests

sozercan (approved):
Thank you! LGTM

PR #346, opened 2023-07-29
feat: validate unknown fields for CR

acpana:
This patch adds validation for unknown fields for a Custom Resource (CR) according to its schema.

This extra validation now requires that we update the test handler and test clients to comply with said validation.

### DIFF
--- a/constraint/pkg/client/crds/validate.go
+++ b/constraint/pkg/client/crds/validate.go
@@ -87,7 +87,7 @@ func ValidateCR(cr *unstructured.Unstructured, crd *apiextensions.CustomResource
 	}
 
 	// validate that there are no unknown fields in the CR
-	// NewStructural assumes the schema is validate
+	// NewStructural assumes the schema is valid
 	structural, err := schema.NewStructural(crd.Spec.Validation.OpenAPIV3Schema)
 	if err != nil {
 		return fmt.Errorf("%w: %v", constraints.ErrInvalidConstraint, err)
--- a/constraint/pkg/client/crds/validate.go
+++ b/constraint/pkg/client/crds/validate.go
@@ -12,6 +12,8 @@ import (
 	"github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	apiextensionsvalidation "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation"
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
+	structuralpruning "k8s.io/apiextensions-apiserver/pkg/apiserver/schema/pruning"
 	"k8s.io/apiextensions-apiserver/pkg/apiserver/validation"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apivalidation "k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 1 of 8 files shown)


constraint/pkg/client/template_client.go
```diff
@@ -59,7 +60,22 @@ func (e *templateClient) ValidateConstraint(constraint *unstructured.Unstructure
...
 	}
 
-	return crds.ValidateCR(constraint, e.crd)
+	if err := crds.ValidateCR(constraint, e.crd); err != nil {
+		return err
+	}
+
+	// validate that there are no unknown fields in the CR
+	// NewStructural assumes the schema is validate
+	structural, err := schema.NewStructural(e.crd.Spec.Validation.OpenAPIV3Schema)
```
maxsmythe, line 69: This should go into the ValidateCR function.

acpana: Moved to `ValidateCR`

maxsmythe (approved):

maxsmythe:
LGTM, may want to un-mark as draft

constraint/pkg/client/crds/validate.go
```diff
@@ -84,5 +86,16 @@ func ValidateCR(cr *unstructured.Unstructured, crd *apiextensions.CustomResource
 		return fmt.Errorf("%w: %v", constraints.ErrSchema, err.ToAggregate())
 	}
 
+	// validate that there are no unknown fields in the CR
+	// NewStructural assumes the schema is validate
```
ritazh, line 90:
```suggestion
	// NewStructural assumes the schema is valid
```

constraint/pkg/client/crds/validate.go
```diff
@@ -84,5 +86,16 @@ func ValidateCR(cr *unstructured.Unstructured, crd *apiextensions.CustomResource
...
+	structural, err := schema.NewStructural(crd.Spec.Validation.OpenAPIV3Schema)
+	if err != nil {
+		return fmt.Errorf("%w: %v", constraints.ErrInvalidConstraint, err)
+	}
+	unknownFields := structuralpruning.PruneWithOptions(cr.DeepCopy().Object, structural, true, schema.UnknownFieldPathOptions{TrackUnknownFieldPaths: true})
+	if len(unknownFields) > 0 {
+		return fmt.Errorf("%w: %v", constraints.ErrInvalidConstraint, fmt.Sprintf("unknown fields: %v", unknownFields))
```
ritazh, line 97: wouldn't this have backward compatibility issues?

acpana: Great question! The k8s api server already does this kind of validation. One can run into this error by trying to apply a constraint with some random field via `kubectl` today. However, `gator` lacks that behavior. For context, this work is being done as part of https://github.com/open-policy-agent/gatekeeper/issues/2863

So, in short, it is not backwards incompatible wrt to the `on cluster` behavior. Wrt to the `off cluster` behavior, mmmaybe but it's not respecting the openapi v3 schema.

@maxsmythe wdy?

maxsmythe: To nit, the API server doesn't reject unknown fields, or raise an error, it silently prunes them. Kubectl raises errors.

I don't think this would be backwards compatibility so much as surfacing undetected bugs in users' code. If users are using legacy schema for parameters, then the behavior is unaltered (since we explicitly allow for unknown fields). If users have unknown fields elsewhere:

* The field may be pruned by the API server
* Kubectl may already be giving them warnings
* They may have spelling errors that would otherwise have gone un-noticed (`enfrocementAction`)

Because we control the schema and semantic meaning of constraint fields (other than parameters, which we've already covered), and because the API server already prunes unknown fields, I don't think previous uses of unrecognized fields could be considered valid inputs subject to backwards compatibility protection.

acpana: thanks for chiming in Max and for making the clear distinctions.

@ritazh does that answer your concern?

ritazh: Yes, thanks for the explanation.

ritazh (approved):
LGTM

PR #372, opened 2023-11-04
chore: upgrade to k8s 1.28.3

maxsmythe:
Upgrades underlying k8s libraries to 1.28.3. This also lets us remove the forked K8s code from the k8s cel driver.


constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
```
ritazh, line 6: Did you mean to downgrade this?

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -197,12 +206,9 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
 			VersionedObject:    request.GetObject(),
 		}
-		parameters, _, err := unstructured.NestedMap(constraint.Object, "spec", "parameters")
-		if err != nil {
-			return nil, err
-		}
-		ctx := context.WithValue(ctx, variablesKey, map[string]interface{}{"params": parameters})
-		response := validator.Validate(ctx, versionedAttr, constraint, celAPI.PerCallLimit)
+
+		// TODO: should namespace be made available, if possible? Generally that context should be present
```
ritazh, line 210: should we open an issue for this to make sure we add ns later?

constraint/pkg/client/drivers/k8scel/temporarydeleteme/compile.go
```diff
@@ -1,285 +0,0 @@
...
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-// forked from k8s.io/apiserver/pkg/admission/plugin/cel/compile.go @ Kubernetes version 1.27.2
-
-//nolint:all do not lint K8s code
-package temporarydeleteme
```
ritazh, line 20: yay for being able to delete this

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
maxsmythe: It looks like that is the version indicated by the makefile:

https://github.com/open-policy-agent/frameworks/blob/2e0cb3d685752d32bfa10fa36f16a126ac68c9de/constraint/Makefile#L39C2-L39C68

constraint/pkg/client/drivers/k8scel/driver.go
maxsmythe: Maybe? Since this is still prototype code and will need to be re-worked for testing etc., was keeping things lightweight.

maxsmythe: LMK if you definitely want to see an issue.

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
acpana: Hm, that is odd -- could we separate fixing this version drift from this PR? Aka do it as a follow up? I can take it on if you don't have cycles for that.

constraint/pkg/client/drivers/k8scel/driver.go
ritazh: it’s just a suggestion, not a PR blocker

ritazh (approved):
LGTM

constraint/config/crds/externaldata.gatekeeper.sh_providers.yaml
ritazh: yea lets address this before bumping it in GK

PR #376, opened 2023-11-14
feat: Add ValidatingAdmissionPolicy generation function

maxsmythe:
Adds the ability to generate a VAP policy definition from a constraint template using the K8sNative engine.

This is an incremental step towards implementing the [Gatekeeper-to-VAP User Flow](https://docs.google.com/document/d/1UELB0dmt10K5jBwETUnSKKQN7qeivOjvi-AvzFTOsHw/edit#heading=h.7yw27nn4g6nn)


constraint/pkg/client/drivers/k8scel/transform/cel_snippets_test.go
```diff
@@ -0,0 +1,733 @@
...
+		return nil, err
+	}
+	var ret interface{}
+	if err := json.Unmarshal(bytes, &ret); err != nil {
+		return nil, err
+	}
+	return ret, nil
+}
+
+func TestMatchKinds(t *testing.T) {
```
acpana, line 72: is it worth having a test that includes a mix of matchers ?

acpana (approved):
excited to see this happening 🎉 

left some minor, non blocking comments

constraint/pkg/client/drivers/k8scel/transform/cel_snippets_test.go
maxsmythe: it is not worth it

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -103,25 +196,34 @@ func (in *Source) GetFailurePolicy() (*admissionv1.FailurePolicyType, error) {
...
-	return nil
+
+	return &out, nil
 }
 
 // ToUnstructured() is a convenience method for converting to unstructured.
-// Intended for testing. It will panic on error. TODO: rename to MustToUnstructured()?
-func (in *Source) ToUnstructured() map[string]interface{} {
+// Intended for testing. It will panic on error.
+func (in *Source) MustToUnstructured() map[string]interface{} {
```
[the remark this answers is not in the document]
maxsmythe: Ease of writing tests

constraint/pkg/client/drivers/k8scel/transform/cel_snippets.go
```diff
@@ -0,0 +1,175 @@
...
+					)
+			)
+		)
+	)
+	`
+)
+
+func MatchExcludedNamespacesGlobV1Alpha1() admissionregistrationv1alpha1.MatchCondition {
+	return admissionregistrationv1alpha1.MatchCondition{
+		Name:       "g8r_match_excluded_namespaces",
```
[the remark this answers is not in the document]
maxsmythe: It cannot change down the road, changing the reserved prefix would be a backwards incompatible change.

constraint/pkg/client/drivers/k8scel/transform/request.go
```diff
@@ -0,0 +1,157 @@
...
+type RequestWrapper struct {
+	ar               *admissionv1.AdmissionRequest
+	object           runtime.Object
+	oldObject        runtime.Object
+	operationOptions runtime.Object
+}
+
+func NewWrapper(req *admissionv1.AdmissionRequest) (*RequestWrapper, error) {
+	var object runtime.Object
+	if len(req.Object.Raw) != 0 {
```
ritazh, line 40: should we return error early for all the len == 0 cases?

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -16,7 +17,8 @@ import (
 const (
 	// Name is the name of the driver.
 	Name           = "K8sNativeValidation"
-	ReservedPrefix = "g8r-"
+	ReservedPrefix = "g8r_"
+	ParamsName     = "params"
```
ritazh, line 21: would be good to add few comments for these new const so we know how they are intended to be used

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -25,18 +27,23 @@ var (
 )
 
 type Validation struct {
-	// A CEL expression. Maps to ValidationAdmissionPolicy's spec.validations.expression
-	Expression        string
-	Message           string
-	MessageExpression string
+	// A CEL expression. Maps to ValidationAdmissionPolicy's spec.validations
```
ritazh, line 30: I think this comment is meant for the `Validation` type so might be good to move it before the type definition.

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -25,18 +27,23 @@ var (
...
 
 type MatchCondition struct {
-	Name string `json:"name"`
-
+	Name       string `json:"name"`
 	Expression string `json:"expression"`
 }
 
+type Variable struct {
+	// A CEL variable definition. Maps to ValidationAdmissionPolicy's spec.variables
```
ritazh, line 42: same, move comment before the type

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -15,8 +16,12 @@ import (
 
 const (
 	// Name is the name of the driver.
-	Name           = "K8sNativeValidation"
-	ReservedPrefix = "g8r-"
+	Name = "K8sNativeValidation"
+	// ReservedPrefix signifies a prefix that no user-defined value (variable, matcher, etc.) is allowed to have.
+	// This gives us the ability to add new variables in the future without worrying about breaking pre-existing templates.
+	ReservedPrefix = "g8r_internal_"
```
sozercan, line 22: Should we rename this to `gatekeeper_...`? Users/contributors might not know what g8r means? Maybe rename K8s to Kubernetes too?

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
```diff
@@ -0,0 +1,59 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TemplateToPolicyDefinition(template *templates.ConstraintTemplate) (*admissionregistrationv1alpha1.ValidatingAdmissionPolicy, error) {
+	source, err := schema.GetSourceFromTemplate(template)
+	if err != nil {
+		return nil, err
+	}
+
+	matchConditions, err := source.GetV1Alpha1MatchConditions()
```
ritazh, line 19: why do we only convert to the alpha one? backward compat? when should we consider beta and v1?

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -65,6 +99,56 @@ func (in *Source) GetMatchConditions() ([]cel.ExpressionAccessor, error) {
...
+			Name:       v.Name,
+			Expression: v.Expression,
+		}
+	}
+
+	return vars, nil
+}
+
+func (in *Source) GetV1Alpha1Variables() ([]admissionv1alpha1.Variable, error) {
+	var variables []admissionv1alpha1.Variable
```
ritazh, line 142: do we not need to validate variables here as well?

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -65,6 +99,56 @@ func (in *Source) GetMatchConditions() ([]cel.ExpressionAccessor, error) {
 	return matchConditions, nil
 }
 
+func (in *Source) GetV1Alpha1MatchConditions() ([]admissionv1alpha1.MatchCondition, error) {
+	var matchConditions []admissionv1alpha1.MatchCondition
```
ritazh, line 103: does this need validation as well?

constraint/pkg/client/drivers/k8scel/transform/cel_snippets_test.go
```diff
@@ -0,0 +1,733 @@
...
+		{
+			name:        "Match against missing namespace",
+			matcher:     nsMatcher{"somename"},
+			shouldMatch: true,
+		},
+		{
+			name:        "Simple match",
+			matcher:     nsMatcher{"somename"},
+			namespace:   ptr.To[string]("somename"),
+			shouldMatch: false,
```
ritazh, line 629: why is shouldMatch false for this case?

constraint/pkg/client/drivers/k8scel/schema/schema.go
maxsmythe: TBH, I'm mostly concerned about brevity/uniqueness. Since this is more about reserving a namespace for us, ideally users never collide with the naming and therefore never need to know the details.

I'm okay with either, but would rather not rename things more than one more time in this PR.

@ritazh preferred name?

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
maxsmythe: I'd say we'd switch once its possible to generate v1beta1 (or greater) VAP resources.

If we wanted to avoid future breaking changes, I could rename this to `TemplateToV1Alpha1PolicyDefinition()` if you prefer?

constraint/pkg/client/drivers/k8scel/transform/cel_snippets_test.go
maxsmythe: b/c its testing excluded namespaces

maxsmythe (commented):
Sorry, for some reason my comments were listed as "pending review", so some of these are older responses (but probably still relevant)

constraint/pkg/client/drivers/k8scel/schema/schema.go
maxsmythe: added

constraint/pkg/client/drivers/k8scel/schema/schema.go
maxsmythe: added

constraint/pkg/client/drivers/k8scel/schema/schema.go
ritazh: I'm ok with either as well. +1 on "gatekeeper_" since g8r is less well known.

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
ritazh: I'm ok keeping it as is.

ritazh (approved):
LGTM

Excited to integrate this!

PR #386, opened 2024-01-06
Create VAP binding objects; update semantic equals to include labels

maxsmythe:
This should be the last remaining object generation work for supporting generating VAP objects.

Next step is to add consideration for the "should be enforced by VAP" labels.


ritazh (approved):
LGTM

PR #393, opened 2024-02-03
Run matchers against both object and oldObject

maxsmythe:
This handles matchers in the case of DELETE. Technically it also handles cases where namespace/name changes, but that should not be possible.


ritazh (approved):
LGTM

PR #400, opened 2024-03-06
Add switch for whether VAP or webhook should enforce constraints

maxsmythe:


constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -168,11 +190,29 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
 		evalStartTime := time.Now()
 		// template name is the lowercase of its kind
-		validator := d.validators[strings.ToLower(constraint.GetKind())]
-		if validator == nil {
+		wrappedValidator := d.validators[strings.ToLower(constraint.GetKind())]
+		if wrappedValidator == nil {
 			return nil, fmt.Errorf("unknown constraint template validator: %s", constraint.GetKind())
 		}
 
+		assumeVAPEnforcementNotDisabled, err := assumeVAPEnforcementWithDefault(constraint, VAPDefaultYes)
```
ritazh, line 198: Why is this defaulted to Yes when this comment above seems to indicate when no label is provided, vap generation should default to no/disabled

> If no value is provided, VAP generation
// is presumed to be disabled and the engine will
// validate ALL policies.

maxsmythe: This replicates the logic Gatekeeper uses. Unwinding the whole decision tree yields:

* If constraint template is undefined, use global default
* If constraint template is disabled, disable
* If constraint template is enabled, fall back to constraint logic
* If constraint is disabled, disable
* If constraint is enabled, enable so long as constraint template is enabled
* If constraint template is undefined, use template value.

Because the if statement is <if constraint enabled> && <if template enabled>, defaulting to true for the constraint means the constraint template determines the outcome.

Here is G8r's way of expressing the the same thing (note the explicit test for "No"):

https://github.com/open-policy-agent/gatekeeper/blob/45e4552595d5fcaf6f66c3d4335aea2c00973760/pkg/controller/constraint/constraint_controller.go#L309-L321

constraint/pkg/client/drivers/k8scel/driver_test.go
```diff
@@ -0,0 +1,474 @@
...
+				Validations: []schema.Validation{
+					{
+						Expression: `object.metadata.name == "unrecognizable-name"`,
+						Message:    "unexpected name",
+					},
+				},
+			}, ptr.To[string](string(VAPDefaultYes))),
+			constraint:         makeConstraint(ptr.To[string](string(VAPDefaultYes))),
+			isAdmissionRequest: true,
+			expectedViolations: true,
```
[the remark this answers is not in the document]
maxsmythe: VAP generation is presumed disabled because no default is provided to the engine. In the case where VAP generation is presumed disabled, the CF is on the hook for enforcing all policies.

This feature is useful for something like the Gator CLI, or any enforcement point where VAP is not available (basically anywhere but k8s admission).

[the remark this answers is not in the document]
ritazh: I see. The difference is whether  vapDefault was set or not. In the case it’s set e.g ptr.To[vapDefault](VAPDefaultNo),  it can be overridden but if not set, it cannot.

constraint/pkg/client/drivers/k8scel/driver.go
ritazh: To match the same behavior as GK, if constraint is not set to disable, this should also return VAPDefaultYes to use template value?
https://github.com/open-policy-agent/frameworks/pull/400/files#diff-5e2ac34b20b681ef5a8747fac1f9563e8c60060f9609e7ee5d10f63a894fd251R297-R299

constraint/pkg/client/drivers/k8scel/driver_test.go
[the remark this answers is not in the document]
maxsmythe: Updated test description to flag that default == nil

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -291,11 +285,12 @@ func assumeVAPEnforcementWithDefault(obj runtime.Object, vapDef vapDefault) (boo
...
 	case VAPDefaultYes:
-		return true, nil
+		return true
 	case VAPDefaultNo:
-		return false, nil
+		return false
+	// on unrecognized value, use the default
 	default:
-		return false, fmt.Errorf("unknown VAP generation label value, %q, should be \"yes\" or \"no\"", shouldGen)
+		return vapDef == VAPDefaultYes
```
ritazh, line 293: In GK, when label is not set to yes or no, we default to parent/CT value, then should this return VAPDefaultYes?

maxsmythe: This is logically equivalent.

If evaluating for a constraint, the call is:

`assumeVAPEnforcementNotDisabled := assumeVAPEnforcementWithDefault(constraint, VAPDefaultYes)`

For a constraint:

* If label is no, return false
* if label is yes/unknown, return true (vapDef == VAPDefaultYes)

For a constraint template

* If no default is provided, assume VAP is disabled, return false
* If label is no, return false
* If label is yes, return true
* If label is unknown, return according to user-provided default (yes => true, no => false)

Line 199: `if isAdmission && assumeVAPEnforcementNotDisabled && wrappedValidator.assumeVAPEnforcement {`

performs the aggregation:

* if not admission,  do not assume VAP handles enforcement (VAP only works for admission) else
* if the constraint has given an explicit "No" (defined by "for a constraint" above returning false), do not assume VAP handles enforcement
* if the constraint template test assumes no VAP enforcement (as defined by "for a constraint template" returning false), do not assume VAP handles enforcement.

TL;DR constraints can only disable enforcement with an explicit "No"

ritazh: > For a constraint:
...
if label is yes/unknown, return true (vapDef == VAPDefaultYes)

ok vapDef is aways VAPDefaultYes for constraint because of 
https://github.com/open-policy-agent/frameworks/pull/400/files#diff-5e2ac34b20b681ef5a8747fac1f9563e8c60060f9609e7ee5d10f63a894fd251R195

ritazh (approved):
LGTM

constraint/pkg/client/drivers/k8scel/driver.go
maxsmythe: Yep!

constraint/pkg/client/drivers/k8scel/driver.go
maxsmythe: resolved in below thread

PR #403, opened 2024-03-13
feat: adding scopedenforcementactions

JaydipGabani:
This PR implements changes for multi ea/ep design.

The gists of the changes are:
- ConstraintToBinding takes a list of enforcement actions instead of determining them (this will be handled by the reconciler code).
- The client knows which enforcement points it can use. When adding constraints for templates, the constraints store precompiled enforcement actions for each known enforcement point.
- One of the client's enforcement points must match the sourceEP in Review for enforcement. If no enforcement points are passed in Review, enforcement applies to all points known to the client.
- Stored constraints are checked against the source of review requests (gator, audit, webhook). Constraints without scopedEnforcementActions are reviewed for all sources.

For gatekeeper changes and CI tests refer to [PR gatekeeper/3321](https://github.com/open-policy-agent/gatekeeper/pull/3321)


constraint/pkg/apis/constraints/apis.go
```diff
@@ -33,6 +43,8 @@ var (
 // to run.
 //
 // Returns an error if spec.enforcementAction is defined and is not a string.
+
+// TODO: change code to evaluate the enforcementAction field when EA is scoped
```
JaydipGabani, line 47: unnecessary TODO, will remove in feedback.

constraint/pkg/client/client_internal_test.go
```diff
@@ -102,7 +102,7 @@ func TestMultiDriverAddTemplate(t *testing.T) {
 			t.Errorf("Driver C has unexpected state: %v", driverC.GetConstraintsForTemplate(anyTemplate))
 		}
 
-		resp, err := client.Review(ctx, handlertest.NewReview("", "foo", "bar"))
+		resp, err := client.Review(ctx, handlertest.NewReview("", "foo", "bar"), "audit.gatekeeper.sh")
 		if err != nil {
```
JaydipGabani, line 106: TODO: use the constant here and everywhere else for enforcementpoint

sozercan (approved):
LGTM

maxsmythe (commented):
Nice! A few comments, but looking good.

constraint/pkg/client/client.go
```diff
@@ -622,7 +622,7 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
 // Review makes sure the provided object satisfies all stored constraints.
 // On error, the responses return value will still be populated so that
 // partial results can be analyzed.
-func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.QueryOpt) (*types.Responses, error) {
+func (c *Client) Review(ctx context.Context, obj interface{}, sourceEP string, opts ...drivers.QueryOpt) (*types.Responses, error) {
```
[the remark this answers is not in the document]
JaydipGabani: @maxsmythe do we want to move `QueryOpt` under client then? How do you imagine passing `QueryOpt` to `review`?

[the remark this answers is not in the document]
maxsmythe: Yeah, client probably needs to own QueryOpt, or it should be in a separate package, since putting it in client may lead to an import cycle.

> How do you imagine passing QueryOpt to review?

I'm a bit confused by this question. Isn't it already being passed to Review()?

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -210,21 +210,29 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
 		// TODO: should namespace be made available, if possible? Generally that context should be present
 		response := validator.Validate(ctx, versionedAttr.GetResource(), versionedAttr, constraint, nil, celAPI.PerCallLimit, nil)
 
-		enforcementAction, found, err := unstructured.NestedString(constraint.Object, "spec", "enforcementAction")
+		enforcementAction, err := apiconstraints.GetEnforcementAction(constraint)
```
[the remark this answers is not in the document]
maxsmythe: Query() should not care about enforcement action. All enforcement-action-processing-logic should live in the client.

Client logic would look like:

* Perform matching as usual and also check to make sure there is an associated enforcement action in order to presume a match (I think you already do this)

* Send a list of constraints-to-be-evaluated to the driver. The driver does not need to know what enforcement points are involved. It does not need to return any enforcement action. The driver can assume that if it has been passed a constraint, that constraint can/should be evaluated.

* Once the client receives results back from the driver, it can re-associate the responses with the enforcement actions that should be taken.

How enforcement actions should be returned from the client is a bit unknown... if we want to support parameters that customize the enforcement action, using a flat list might not work, but the CF interface is still a prototype, so not too worried there. In any case, that still doesn't require the driver to know about EAs/EPs.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
JaydipGabani: Yip! but as a part of refactor we will not be passing the same to `Review`. I found a way to set those options and pass it to `review`.

JaydipGabani:
@maxsmythe Sorry for the delay. I have addressed all the comments and implemented the feedback. PTAL.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +56,123 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+// GetEnforcementActionsForEP returns a map of enforcement actions for enforcement points.
+func GetEnforcementActionsForEP(constraint *unstructured.Unstructured, eps []string) (map[string]map[string]bool, error) {
+	actionsForEPs := make(map[string]map[string]bool)
+
+	for _, enforcementPoint := range eps {
+		actionsForEPs[enforcementPoint] = make(map[string]bool)
+	}
+
+	// Access the scopedEnforcementAction field
+	scopedActions, found, err := getNestedFieldAsArray(constraint.Object, "spec", "scopedEnforcementActions")
```
maxsmythe, line 73: It might be helpful to simply define a `ScopedEnforcementActions` struct and convert the JSON to the struct via DefaultUnstructuredConverter (example):

https://github.com/open-policy-agent/frameworks/blob/5446e9edab0b7c9177335e6ef0e286e2a435ff4f/constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go#L113

That way:

* we have a place to point to for the expected schema of scopedEnforcementActions
* You don't need to deal with getNested<thing> commands/error processing/downstream conversions

constraint/pkg/client/client.go
```diff
@@ -622,7 +625,26 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
 // Review makes sure the provided object satisfies all stored constraints.
 // On error, the responses return value will still be populated so that
 // partial results can be analyzed.
-func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.QueryOpt) (*types.Responses, error) {
+func (c *Client) Review(ctx context.Context, obj interface{}, sourceEPs ...string) (*types.Responses, error) {
```
maxsmythe, line 628: Review should still take a variadic argument of opt... they just shouldn't be QueryOpts anymore (ReviewOpts?)

basically:

* Move QueryOpts to its own package separate from client & drivers (so both can import), rename to ReviewOpt
* Add a new option to specify source EPs
* Maybe separate out query opts that should be passed to drivers as their own class to avoid blurring abstractions.

The reason you don't want to use the variadic argument purely for EPs is that then there is nowhere to make backwards-compatible changes.

constraint/pkg/client/client_opts.go
```diff
@@ -72,3 +79,36 @@ func IgnoreNoReferentialDriverWarning(ignore bool) Opt {
...
 }
+
+func EnforcementPoints(eps ...string) Opt {
+	return func(client *Client) error {
+		client.enforcementPoints = eps
+		return nil
+	}
+}
+
+func SetDriverQueryOpt(sourceEP string, opt ...drivers.QueryOpt) {
```
maxsmythe, line 90: This doesn't quite work, since query options should be expressible on a per-call-to-review basis (see alternate design I suggested in a comment on the function signature for Review())

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
```diff
@@ -79,14 +79,42 @@ func ConstraintToBinding(constraint *unstructured.Unstructured) (*admissionregis
...
-		enforcementAction = admissionregistrationv1beta1.Warn
-	default:
-		return nil, fmt.Errorf("%w: unrecognized enforcement action %s, must be `warn` or `deny`", ErrBadEnforcementAction, enforcementActionStr)
+	actions := []string{}
+	if apiconstraints.IsEnforcementActionScoped(enforcementActionStr) {
+		actionsForEP, err := apiconstraints.GetEnforcementActionsForEP(constraint, []string{apiconstraints.WebhookEnforcementPoint})
+		if err != nil {
+			return nil, err
+		}
+		if len(actionsForEP[apiconstraints.WebhookEnforcementPoint]) == 0 {
```
maxsmythe, line 88: Should VAP be its own enforcement point?

maxsmythe (changes requested):
No worries! Sorry for the delay on my end. Getting close!

constraint/pkg/apis/constraints/apis.go
```diff
@@ -17,6 +17,17 @@ const (
...
+	EnforcementActionScoped = "scoped"
+
+	// WebhookEnforcementPoint is the enforcement point for admission.
+	WebhookEnforcementPoint = "validation.k8s.io"
+
+	// AuditEnforcementPoint is the enforcement point for audit.
+	AuditEnforcementPoint = "audit.gatekeeper.sh"
+
+	// GatorEnforcementPoint is the enforcement point for gator cli.
+	GatorEnforcementPoint = "gator.gatekeeper.sh"
```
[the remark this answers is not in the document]
JaydipGabani: Only `WebhookEnforcementPoint` constant is used to retrieve `enforcementAction` to create respective VAP with - https://github.com/open-policy-agent/frameworks/pull/403/files#diff-ac56566503d2145ac94cc1edbd630db0a048436f9c916956fd9ea35245636989R84 - audit and gator ep constants are only used in tests. So it is possible to have custom eps while integrating CF.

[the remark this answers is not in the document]
JaydipGabani: For the VAP if we would need one predefined EP to properly create VAPB from constraint.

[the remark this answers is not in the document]
ritazh: One use for EP having strong types is if we plan to do validations against allowed actions for each EP. e.g. Deny, warn actions for the audit EP doesn't really make sense.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +56,123 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+	}
+	return nil, fmt.Errorf("value must be a map[string]interface{}")
+}
+
+// Helper function to convert a value to a []map[string]interface{}.
+func convertToMapSlice(value interface{}) ([]map[string]interface{}, error) {
+	if arr, ok := value.([]interface{}); ok {
+		result := make([]map[string]interface{}, 0, len(arr))
+		for _, v := range arr {
+			if m, ok := v.(map[string]interface{}); ok {
```
ritazh, line 171: This currently skips elements that have values that are not of type map[string]interface{}. Is that intentional? do we need to return an error or log error?

constraint/pkg/client/client.go
```diff
@@ -622,7 +625,26 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
 // Review makes sure the provided object satisfies all stored constraints.
```
ritazh, line 625: please update the comment to reflect the change

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
ritazh: that makes sense and the actions will be mapped as: deny, warn, audit as part of the generated vap resource.

constraint/pkg/apis/constraints/apis.go
JaydipGabani: Updated the code as per max's comment here - https://github.com/open-policy-agent/frameworks/pull/403/#discussion_r1558699319

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
JaydipGabani: does `validation.gatekeeper.sh -> gatekeeper webhook & admission.k8s.io -> VAP` work?

constraint/pkg/apis/constraints/apis.go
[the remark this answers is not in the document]
JaydipGabani: But then `audit` action with `gator` or `gk-webhook` enforcementPoint doesnt match up.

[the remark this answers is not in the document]
maxsmythe: IMO audit is definitely an EP.

IIRC we suggested that EPs could do an analysis and report on the status of the constraint (or whatever status reporting mechanism the EP has) if there was some kind of invalid enforcement action?

The problem with Validating all known EPs or all acceptable actions for EPs are:

- CF can only hard code that logic for EPs it knows about, which limits its ability to be re-used
- Validating that all EPs are recognized assumes that the thing doing the validating knows about all available EPs. This may not be true (e.g. Gatekeeper knows nothing about terraform validator)
- An EP validating that it understands the actions it itself is intended to take makes total sense

IMO I'd avoid creating these constants here, since it implies that they can be used for things like validation. Definitely makes sense to define a special EP inside the VAP driver (whether that is  admission.k8s.io or something else).

Interesting side note: this dovetails nicely with the idea that the VAP driver probably should live in G8r rather than in the CF (that idea came up in other PRs)

constraint/pkg/client/drivers/k8scel/transform/make_vap_objects.go
maxsmythe: Maybe? Probably best to bikeshed on naming at the community meeting.  Admission webhooks are their own concept in K8s, IIRC.

vap.k8s.io?

constraint/pkg/apis/constraints/apis.go
maxsmythe: We probably want to give an error if there is unrecognized data

JaydipGabani: Yeah, we are now returning errors on unrecognized data.

sozercan (approved):
LGTM

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +60,86 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+
+	for _, scopedEnforcementAction := range scopedEnforcementActions {
+		scopedEA := &ScopedEnforcementAction{}
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(scopedEnforcementAction, scopedEA); err != nil {
+			return nil, err
+		}
+
+		// Iterate over enforcementPoints
+		for _, enforcementPoint := range scopedEA.EnforcementPoints {
+			if _, ok := actionsForEPs[enforcementPoint.Name]; !ok && enforcementPoint.Name != "*" {
```
ritazh, line 101: Please make sure `actionsForEPs` is not nil. Accessing nil maps will result in a runtime panic.

ritazh: we should make enforcementPoint.Name all lower case to avoid case sensitivity issues

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +60,86 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+				continue
+			}
+			switch enforcementPoint.Name {
+			case "*":
+				for _, ep := range eps {
+					actionsForEPs[ep][scopedEA.Action] = true
```
ritazh, line 107: also need to make sure each scopedEA.Action is lower case to avoid case sensitivity issues.

constraint/pkg/apis/constraints/apis.go
ritazh: I don't see any error returned when v is not of type map[string]interface{}
> if m, ok := v.(map[string]interface{}); ok {

nothing is checking the !ok

constraint/pkg/client/client.go
```diff
@@ -619,10 +623,32 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
...
+func (c *Client) Review(ctx context.Context, obj interface{}, opts ...reviews.ReviewOpt) (*types.Responses, error) {
+	var eps []string
+	cfg := &reviews.ReviewCfg{}
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	if cfg.SourceEP != "" {
+		for _, ep := range c.enforcementPoints {
+			if cfg.SourceEP == ep {
```
ritazh, line 638: ensure case sensitivity

constraint/pkg/client/client.go
```diff
@@ -682,6 +712,10 @@ func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.Qu
 			continue
 		}
 
+		for i := range resp.Results {
+			resp.Results[i].EnforcementAction = enforcementActionsByTarget[target][resp.Results[i].Constraint.GetName()]
```
ritazh, line 716: ensure enforcementActionsByTarget is not nil, and enforcementActionsByTarget[target] is not nil

ritazh: add tests for these cases

constraint/pkg/client/constraint_client.go
```diff
@@ -16,24 +16,34 @@ type constraintClient struct {
...
-func (c *constraintClient) matches(target string, review interface{}) *constraintMatchResult {
+func (c *constraintClient) matches(target string, review interface{}, sourceEPs ...string) *constraintMatchResult {
 	matcher, found := c.matchers[target]
 	if !found {
 		return nil
 	}
 
+	var actions []string
+	for _, ep := range sourceEPs {
+		if acts, found := c.enforcementActionsForEP[ep]; found {
```
ritazh, line 38: check case sensitivity and spaces

constraint/pkg/client/drivers/to_result.go
```diff
@@ -74,20 +73,10 @@ func ToResult(constraints map[ConstraintKey]*unstructured.Unstructured, r rego.R
...
+
 	// DeepCopy the result so we don't leak internal state.
 	result.Constraint = constraint.DeepCopy()
 
-	enforcementAction, found, err := unstructured.NestedString(constraint.Object, "spec", "enforcementAction")
-	if err != nil {
-		return nil, err
-	}
-	if !found {
-		enforcementAction = apiconstraints.EnforcementActionDeny
```
ritazh, line 87: where do we test if `enforcementAction` and `scopedEnforcementActions` are both not provided

constraint/pkg/client/client.go
JaydipGabani: > ensure enforcementActionsByTarget is not nil, and enforcementActionsByTarget[target] is not nil

This will never be the case as `enforcementActionsByTarget` is initialized by `make()` and without any `target` this line will not get executed at all. I added a check to verify the entry in the map to make sure just in case

constraint/pkg/apis/constraints/apis.go
JaydipGabani: refactored the function to return proper error.

constraint/pkg/client/client.go
JaydipGabani: Storing `sourceEP` and `c.enforcementPoints` in lowercase at initialisation to avoid multiple `ToLower()` calls while checking for values.

constraint/pkg/client/constraint_client.go
JaydipGabani: Storing the initial value in lowercase so this should already be in lowercase.

constraint/pkg/client/drivers/to_result.go
JaydipGabani: This test is here - [e2e_test.go#L560](https://github.com/open-policy-agent/frameworks/pull/403/files#diff-7ba8c60c42c538a187e71a73932b6a6a7535ad0a5e8c06e3c50e8b02fc772343R560)

constraint/pkg/client/e2e_test.go
```diff
@@ -478,10 +661,12 @@ func TestClient_Review_Details(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	fmt.Println(responses)
```
ritazh, line 664: is this for debugging?

JaydipGabani: good catch! I will remove this.

constraint/pkg/apis/constraints/apis_test.go
```diff
@@ -0,0 +1,212 @@
...
+										"name": "*",
+									},
+								},
+								"action": "deny",
+							},
+						},
+					},
+				},
+			},
+			expected: map[string]map[string]bool{
```
ritazh, line 102: why is AuditEnforcementPoint excluded from expected and eps?

ritazh: nvm i see eps is an input, not expected result

constraint/pkg/client/client.go
```diff
@@ -619,10 +623,55 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
...
+			}
+			// If the specified source enforcement point matches the current enforcement point, add it to the list
+			if cfg.SourceEP == ep {
+				eps = append(eps, ep)
+				break // Exit the loop since the matching enforcement point is found
+			}
+		}
+		// If enforceAll is true, add the source enforcement point to the list of enforcement points
+		if enforceAll {
+			eps = append(eps, cfg.SourceEP)
```
ritazh, line 657: what if cfg.SourceEP is already appended?

constraint/pkg/client/client.go
```diff
@@ -619,10 +623,55 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
...
+	var eps []string
+	// Create a new ReviewCfg instance
+	cfg := &reviews.ReviewCfg{}
+	// Apply each option to the ReviewCfg instance
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	// Check if a source enforcement point is specified in the configuration
+	if cfg.SourceEP != "" {
```
ritazh, line 640: who sets cfg.SourceEP?

constraint/pkg/client/client.go
ritazh: nvm I see it in GK

constraint/pkg/client/client.go
ritazh: nvm I see it in GK

constraint/pkg/client/constraint_client.go
```diff
@@ -16,24 +17,58 @@ type constraintClient struct {
...
+	enforcementActions := make(map[string]bool)
+	if apiconstraints.IsEnforcementActionScoped(c.enforcementAction) {
+		// Initialize a map to track unique enforcement actions
+		// Iterate over the provided source enforcement points (EPs)
+		for _, ep := range sourceEPs {
+			var actions []string
+			// Check if there are predefined actions for the current EP
+			if acts, found := c.enforcementActionsForEP[ep]; found {
+				actions = acts // Use the predefined actions if found
+			} else if ep == "*" {
```
ritazh, line 50: should this be checked first and exit early if it's `*` since actions is already set.

ritazh: the current implementation will override actions if another ep is evaluated after *

constraint/pkg/client/constraint_client.go
```diff
@@ -16,24 +17,58 @@ type constraintClient struct {
...
 
-func (c *constraintClient) matches(target string, review interface{}) *constraintMatchResult {
+func (c *constraintClient) matches(target string, review interface{}, sourceEPs ...string) *constraintMatchResult {
 	matcher, found := c.matchers[target]
 	if !found {
 		return nil
 	}
 
+	enforcementActions := make(map[string]bool)
+	if apiconstraints.IsEnforcementActionScoped(c.enforcementAction) {
```
ritazh, line 42: if IsEnforcementActionScoped == false is not handled

constraint/pkg/client/constraint_client.go
```diff
@@ -16,24 +17,58 @@ type constraintClient struct {
...
+			for _, act := range actions {
+				enforcementActions[act] = true
+			}
+		}
+
+	}
+	// If no enforcement actions are found, return nil
+	if len(enforcementActions) == 0 && apiconstraints.IsEnforcementActionScoped(c.enforcementAction) {
+		return nil
+	}
```
ritazh, line 66: add a comment for when IsEnforcementActionScoped == false, enforcementActions implicitly would be > 0

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -200,20 +200,12 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
 		// TODO: should namespace be made available, if possible? Generally that context should be present
 		response := validator.Validate(ctx, versionedAttr.GetResource(), versionedAttr, constraint, nil, celAPI.PerCallLimit, nil)
 
-		enforcementAction, found, err := unstructured.NestedString(constraint.Object, "spec", "enforcementAction")
-		if err != nil {
-			return nil, err
-		}
-		if !found {
-			enforcementAction = apiconstraints.EnforcementActionDeny
```
ritazh, line 208: We should not remove this for non-scope EA, this is still needed for backward compat.

constraint/pkg/client/drivers/to_result.go
ritazh, line 87: why do we need to remove this? we still need it for non-scope EA.

JaydipGabani: `enforementAction` and `scopedEnforcementActions` are set [here ](https://github.com/open-policy-agent/frameworks/pull/403/files/2428b2195db8b85fd45f55d16fc1559c2ad00203#diff-2b54391e78f46b40ac71789b492e315a83cbd6b15716c000dcd4c5915d907802R741) - once driver is done with `query` and returns responses.

constraint/pkg/client/drivers/k8scel/driver.go
JaydipGabani: enforementAction and scopedEnforcementActions are set [here ](https://github.com/open-policy-agent/frameworks/pull/403/files/2428b2195db8b85fd45f55d16fc1559c2ad00203#diff-2b54391e78f46b40ac71789b492e315a83cbd6b15716c000dcd4c5915d907802R741)- once driver is done with query and returns responses.

constraint/pkg/client/e2e_test.go
```diff
@@ -34,6 +34,8 @@ func TestClient_Review(t *testing.T) {
 		constraints []*unstructured.Unstructured
 		inventory   []*handlertest.Object
 		toReview    interface{}
+		sourceEP    string
+		clientEPs   []string
```
ritazh, line 38: can you pls add a comment for what is sourceEP and what is clientEPs

constraint/pkg/client/drivers/to_result.go
ritazh: where do you set if enforcementAction is not found, set to deny?

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +64,101 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+	if arr, ok := value.([]interface{}); ok {
+		for _, v := range arr {
+			if m, ok := v.(map[string]interface{}); ok {
+				scopedEA := &ScopedEnforcementAction{}
+				if err := runtime.DefaultUnstructuredConverter.FromUnstructured(m, scopedEA); err != nil {
+					return nil, err
+				}
+				result = append(result, *scopedEA)
+			} else {
+				return nil, fmt.Errorf("scopedEnforcementActions value must be a []scopedEnforcementAction{action: string, enforcementPoints: []EnforcementPoint{name: string}}")
```
ritazh, line 158: You can consider checking these conditions first, then return. this will reduce the number of elses you need in the code.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +64,101 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+			}
+		}
+		return result, nil
+	}
+	return nil, fmt.Errorf("scopedEnforcementActions value must be a []scopedEnforcementAction{action: string, enforcementPoints: []EnforcementPoint{name: string}}")
```
ritazh, line 163: same

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +64,101 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+				scopedEA := &ScopedEnforcementAction{}
+				if err := runtime.DefaultUnstructuredConverter.FromUnstructured(m, scopedEA); err != nil {
+					return nil, err
+				}
+				result = append(result, *scopedEA)
+			} else {
+				return nil, fmt.Errorf("scopedEnforcementActions value must be a []scopedEnforcementAction{action: string, enforcementPoints: []EnforcementPoint{name: string}}")
+			}
+		}
+		return result, nil
```
ritazh, line 161: same

constraint/pkg/client/drivers/to_result.go
JaydipGabani: It gets populated while adding constraint - [here](https://github.com/open-policy-agent/frameworks/pull/403/files/2428b2195db8b85fd45f55d16fc1559c2ad00203#diff-90ee2c69c3c3a9b661e6e71944b7cde7a50a126a49b694afedb30a8a4f50da81R102)

constraint/pkg/client/client.go
JaydipGabani: Fixed it

constraint/pkg/client/constraint_client.go
JaydipGabani: fixed it.

constraint/pkg/client/constraint_client.go
JaydipGabani: Added comment.

constraint/pkg/client/template_client.go
```diff
@@ -95,10 +95,46 @@ func (e *templateClient) Update(templ *templates.ConstraintTemplate, crd *apiext
...
+			}
+			// Initialize a slice to hold actions for the current EP, with capacity equal to the number of actions
+			enforcementActionsForEP[ep] = make([]string, 0, len(actions))
+			// Add each action to the slice for the current EP
+			for action := range actions {
+				enforcementActionsForEP[ep] = append(enforcementActionsForEP[ep], action)
+			}
+		}
+	} else {
+		// If the enforcement action is not scoped, or if client does not specify EPs, apply the action to all EPs
```
maxsmythe, line 127: The client must always specify EPs

constraint/pkg/client/template_client.go
```diff
@@ -95,10 +95,46 @@ func (e *templateClient) Update(templ *templates.ConstraintTemplate, crd *apiext
...
+		if len(enforcementPoints) == 0 {
+			enforcementPoints = []string{"*"} // Use "*" to represent all EPs
```
maxsmythe, line 129: This is not a place where we should allow `*`, per other comments.

constraint/pkg/client/template_client.go
```diff
@@ -95,10 +95,46 @@ func (e *templateClient) Update(templ *templates.ConstraintTemplate, crd *apiext
...
+		}
+		// Iterate over the enforcement points to set the enforcement action for each
+		for _, ep := range enforcementPoints {
+			// Initialize a slice with capacity 1 for the enforcement action
+			enforcementActionsForEP[ep] = make([]string, 0, 1)
```
maxsmythe, line 134: no need to initialize length/capacity

maxsmythe (changes requested):

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +64,101 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
 
 	return action, nil
 }
+
+func IsEnforcementActionScoped(action string) bool {
+	return strings.EqualFold(action, EnforcementActionScoped)
```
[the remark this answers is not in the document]
JaydipGabani: @maxsmythe were you meaning to refer to some other codeblock? I tested the switch in the code you shared. I do not think it checks for case sensitivity.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -45,3 +64,101 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+		return nil, fmt.Errorf("%w: %w", ErrInvalidConstraint, err)
+	}
+
+	// Flag to indicate if all enforcement points should be enforced
+	enforceAll := false
+	// Initialize a map to hold enforcement actions for each enforcement point
+	actionsForEPs := make(map[string]map[string]bool)
+	// Populate the actionsForEPs map with enforcement points from eps, initializing their action maps
+	for _, enforcementPoint := range eps {
+		if enforcementPoint == AllEnforcementPoints {
```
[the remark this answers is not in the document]
JaydipGabani: I was considering that `reviewopts` and `client` would have `*` to support all enforcement points and that made the whole implementation complex. I reverted to your suggestion wildcard in reviewops and client. So not there is no reson that AllEnforcementPoints would be passed to this function.

constraint/pkg/client/client.go
```diff
@@ -619,10 +623,55 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
...
-// Review makes sure the provided object satisfies all stored constraints.
+// Review makes sure the provided object satisfies constraints applicable for specific enforcement points.
 // On error, the responses return value will still be populated so that
 // partial results can be analyzed.
-func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.QueryOpt) (*types.Responses, error) {
+func (c *Client) Review(ctx context.Context, obj interface{}, opts ...reviews.ReviewOpt) (*types.Responses, error) {
+	// Initialize an empty slice for enforcement points
+	var eps []string
+	// Create a new ReviewCfg instance
+	cfg := &reviews.ReviewCfg{}
```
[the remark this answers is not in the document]
JaydipGabani: Yeah, that was always the case. Now returning error when client does not have EP instead of being calculative about it and enforcing everything.

constraint/pkg/apis/constraints/apis.go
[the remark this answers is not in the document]
ritazh: @JaydipGabani From the code Max linked, it means we should preserve the case sensitivity of enforcementAction instead of converting it to all lowercase. Currently, only these are expectedhttps://github.com/open-policy-agent/gatekeeper/blob/master/pkg/util/enforcement_action.go#L14-L18 so unless users pass in `deny`, `dryrun`, `warn`, `scoped` (new one that we should add to that list), we should set it as `unrecognized`.

constraint/pkg/client/constraint_client.go
ritazh: The comment is about when enforcement action is scoped. Please add one for when it is NOT scoped.

If enforcement action is not scoped, then enforcementAction is passed directly to the results.

JaydipGabani: Added comment for `enforcementAction` and `actions` getting passed to results for when, action is not scoped or there is a matching enforcement point for which actions need to be taken.

constraint/pkg/apis/constraints/apis.go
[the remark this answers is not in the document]
JaydipGabani: Checking for case sensitve matching for actions now.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -40,8 +66,102 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+	}
+	for _, scopedEA := range scopedEnforcementActions {
+		for _, enforcementPoint := range scopedEA.EnforcementPoints {
+			epName := strings.ToLower(enforcementPoint.Name)
+			var action string
+			switch scopedEA.Action {
+			case string(Warn), string(Dryrun):
+				action = scopedEA.Action
+			default:
+				action = string(Deny)
```
ritazh, line 106: shouldnt we return error? we cant assume non valid values just all default to deny. we can only set to deny when enforcementAction is not provided.

ritazh: https://github.com/open-policy-agent/gatekeeper/blob/bd96c5263523be54642dcb4a88c2a9e502a74ea7/pkg/webhook/policy.go#L300-L309 this code assumes not dryrun, not warn, means deny because it's processing the response returned by the driver. but we should not make that assumption here. 

https://github.com/open-policy-agent/gatekeeper/blob/bd96c5263523be54642dcb4a88c2a9e502a74ea7/pkg/webhook/policy.go should also be updated to include scope otherwise, it will default to deny.

JaydipGabani: Right now the behavior is, if someone turns off the `disable-enforcementaction-validation` flag on gatekeeper, then constraints can be created with whatever enforcementAction (`enforcementAction: fake`) - [Link](https://github.com/open-policy-agent/gatekeeper/blob/bd96c5263523be54642dcb4a88c2a9e502a74ea7/pkg/webhook/policy.go#L415). In this case, constraint gets added to the template client regardless of the "valid/invalid" enforcement action. Keeping the behavior same IMO we should just pass whatever scoped enforcement action is in the constraint without validating it when adding the constraint to template client. The mechanism to handle this already exists in gatekeeper when we receive responses in gatekeeper with "unrecognised" actions.

By default, we are validationg enforcement action and scoped enforcement actions to reject constraint with not valid enforcement action. `enforcementAction: deny, dryrun, warn, scoped` and `scopedEnforcementActions[].action: denym dryrun, warn`

 So should I remove this block altogether then? I dont think we need this check here.

JaydipGabani: > https://github.com/open-policy-agent/gatekeeper/blob/bd96c5263523be54642dcb4a88c2a9e502a74ea7/pkg/webhook/policy.go should also be updated to include scope otherwise, it will default to deny.

I think we should keep this as is because when response include `scoped` enforcement action, the implementation looks at `scoped` enforcement actions. https://github.com/open-policy-agent/gatekeeper/pull/3321/files#diff-d3c8e52b699d3596d1057b82890a2ed8624445f1ee8f6d3d8a532e13766b99a8R261

maxsmythe: Rita is correct, the code should be removed.

The fact that constraints can be created with arbitrary enforcement actions is intentional. This is because we cannot know, and should not presume to know, what enforcement actions are available to any EPs.

We should not validate enforcement actions and just take them as we come.

If EPs want to perform additional validation (e.g. Gatekeeper raising a warning/reporting via `status` if it is on the hook for taking an action it does not recognize), that's fine, but CF is not equipped to do that. At best it can provide helper functions to allow EPs to implement that extra porcelain.

Assuming unrecognized actions == deny is not a good assumption to make. Example: if Gatekeeper adds a PubSub action, the CF would rewrite it as deny.

maxsmythe: Also, we should not return an error, as I alluded to earlier, CF does not and should not know the full set of valid enforcement actions, it just forwards them to EPs, who are responsible for interpreting them.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: Where is it that an error is returned if a client does not have EPs?

constraint/pkg/client/constraint_client.go
```diff
@@ -16,24 +17,58 @@ type constraintClient struct {
...
 	matcher, found := c.matchers[target]
 	if !found {
 		return nil
 	}
 
+	enforcementActions := make(map[string]bool)
+	if apiconstraints.IsEnforcementActionScoped(c.enforcementAction) {
+		for _, ep := range enforcementPoints {
+			var actions []string
+			if ep == apiconstraints.AllEnforcementPoints {
```
maxsmythe, line 46: `*` cannot be provided to `matches()`, calls to `matches()` must provide a specific enforcement point.

Remember, `*` only makes sense when being specified on the constraint.

constraint/pkg/client/client.go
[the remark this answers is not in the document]
maxsmythe: NVM, found it

constraint/pkg/client/constraint_client.go
JaydipGabani: Agreed. sorry for missing this. FIxed it.

constraint/pkg/client/client.go
```diff
@@ -619,10 +623,27 @@ func (c *Client) RemoveData(ctx context.Context, data interface{}) (*types.Respo
...
 // On error, the responses return value will still be populated so that
 // partial results can be analyzed.
-func (c *Client) Review(ctx context.Context, obj interface{}, opts ...drivers.QueryOpt) (*types.Responses, error) {
+func (c *Client) Review(ctx context.Context, obj interface{}, opts ...reviews.ReviewOpt) (*types.Responses, error) {
+	var eps []string
+	cfg := &reviews.ReviewCfg{}
+	for _, opt := range opts {
+		opt(cfg)
+	}
+	if cfg.EnforcementPoint == "" {
```
ritazh, line 635: This is the enforcementPoint under scopedEnforcementActions right?

maxsmythe (approved):
Code LGTM, have not reviewed tests, but I defer to Rita there.

constraint/pkg/client/client.go
JaydipGabani: This is enforcement point passed in `review.opts`.

constraint/pkg/apis/constraints/apis.go
```diff
@@ -40,8 +65,95 @@ func GetEnforcementAction(constraint *unstructured.Unstructured) (string, error)
...
+		return nil, fmt.Errorf("%w: %w", ErrInvalidConstraint, err)
+	}
+
+	enforcementPointsToActionsMap := make(map[string]map[string]bool)
+	for _, ep := range eps {
+		enforcementPointsToActionsMap[ep] = make(map[string]bool)
+	}
+	for _, scopedEA := range scopedEnforcementActions {
+		for _, enforcementPoint := range scopedEA.EnforcementPoints {
+			epName := strings.ToLower(enforcementPoint.Name)
```
ritazh, line 99: I don't think we want to do this here since GetEnforcementAction today just passes the value thru as is. If it's not a match compare to the supported EPs, then GK will show it as `unrecognized`.

constraint/pkg/client/client_opts.go
```diff
@@ -72,3 +73,13 @@ func IgnoreNoReferentialDriverWarning(ignore bool) Opt {
 		return nil
 	}
 }
+
+func EnforcementPoints(eps ...string) Opt {
+	return func(client *Client) error {
+		for i, ep := range eps {
+			eps[i] = strings.ToLower(ep)
```
ritazh, line 80: same here

constraint/pkg/apis/constraints/apis.go
ritazh: Since EP can be any string, my preference would be to not converting it to lowercase to preserve it. how do others feel? @maxsmythe @sozercan ?

maxsmythe: I'm down with preserving case

ritazh (approved):
LGTM

PR #411, opened 2024-04-08
feat: llm engine driver

sozercan:
adds a new driver for an llm policy engine


JaydipGabani (approved):

constraint/pkg/client/drivers/llm/driver.go
```diff
@@ -0,0 +1,266 @@
...
+	source, err := llmSchema.GetSourceFromTemplate(ct)
+	if err != nil {
+		return err
+	}
+
+	prompt, err := source.GetPrompt()
+	if err != nil {
+		return err
+	}
+	if prompt == "" {
```
ritazh, line 75: other than empty string, are there other prompts that we may want to prevent? could any prompts be used to exploit or manipulate the system?

PR #427, opened 2024-04-30
chore: adding generateVAP field on template

JaydipGabani:
Replacing `use-vap` annotation by `GenerateVAP` field on templates.


constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
```diff
@@ -79,6 +79,9 @@ spec:
                             description: 'The engine used to evaluate the code. Example:
                               "Rego". Required.'
                             type: string
```
maxsmythe, line 81: `generateVAP` should be a property of the VAP engine, not the CT schema. It should be a member of `source` for the k8snative driver.

Otherwise, we couple CTs to VAP, and only the VAP engine is intended to generate VAP objects anyway

maxsmythe (changes requested):
left comment about where the new field should belong.

constraint/config/crds/templates.gatekeeper.sh_constrainttemplates.yaml
JaydipGabani: @maxsmythe good point, moved `generateVAP` field to `source` for k8snative driver.

constraint/pkg/apis/templates/v1/constrainttemplate_types.go
```diff
@@ -72,6 +72,7 @@ type Code struct {
 	// +kubebuilder:validation:Required
 	Engine string `json:"engine"`
 
+
```
maxsmythe, line 75: nit: extra space

constraint/pkg/client/drivers/k8scel/args.go
```diff
@@ -21,16 +13,15 @@ func GatherStats() Arg {
 }
 
 // VAPGenerationDefault sets the expected default
-// value of the `gatekeeper.sh/use-vap` label.
+// value for generateVAP field.
```
maxsmythe, line 16: nit: "for the"

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -193,11 +194,9 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
 			return nil, fmt.Errorf("unknown constraint template validator: %s", constraint.GetKind())
 		}
 
-		assumeVAPEnforcementNotDisabled := assumeVAPEnforcementWithDefault(constraint, VAPDefaultYes)
```
maxsmythe, line 196: This is going away because VAP enforcement is merely another enforcement point?

maxsmythe (commented):
Looking good!

One nit and one clarifying question. Apologies for the slow turnaround.

constraint/pkg/client/drivers/k8scel/driver.go
JaydipGabani: This is a check to see if we have created VAPB for this constaint or not. We depend on int-tree VAPB if it exists, else we use gk webhook for enforcement. 

You are correct, now we do not have `none` option to disable VAP for template/constraint combo, rather it would be through `scopedEnforcementAction` if user wants to disable it.

maxsmythe: SGTM

constraint/Makefile
```diff
@@ -42,7 +42,7 @@ gen-dependencies:
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests:
-	controller-gen \
+	/home/funkydev/go/bin/controller-gen \
```
maxsmythe, line 45: Need to remove `/home/funkydev` :p

maxsmythe (commented):
LGTM after last nit is addressed

constraint/Makefile
JaydipGabani: My bad. Removed the path.

maxsmythe (approved):
LGTM

ritazh:
@JaydipGabani Can you please update the [design doc](https://docs.google.com/document/d/1UELB0dmt10K5jBwETUnSKKQN7qeivOjvi-AvzFTOsHw/edit#heading=h.7yw27nn4g6nn) with the latest decision and timestamp?

From Slack thread:
> vap-enforcement flag to define a global behavior when the driver field on CT is missing -
do not generate when field is missing for alpha/beta
generate when field is missing for GA
C inherits the behavior from CT.
When vap/vapb generation is disabled, and VAP 
ep is included in scopedEA we error out on C saying VAPB is not enabled for this C on admission.

constraint/pkg/client/drivers/k8scel/driver_test.go
```diff
@@ -37,38 +37,29 @@ func makeTemplateWithSource(source *schema.Source, vapGenerationVal *string) *te
...
 				Message:    "Always true",
 			},
 		},
-	}, vapGenerationVal)
+		GenerateVAP: vapGenerationVal,
+	})
 }
 
-func makeConstraint(vapGenerationVal *string) *unstructured.Unstructured {
+func makeConstraint() *unstructured.Unstructured {
```
ritazh, line 55: I forgot how can users override at the constraint level? @JaydipGabani  can you pls add this here and to the design doc?

constraint/pkg/client/drivers/k8scel/driver_test.go
```diff
@@ -285,42 +276,13 @@ func TestValidation(t *testing.T) {
...
-			template: makeTemplateWithSource(&schema.Source{
-				Validations: []schema.Validation{
-					{
-						Expression: `object.metadata.name == "unrecognizable-name"`,
-						Message:    "unexpected name",
-					},
-				},
-			}, nil),
-			constraint:         makeConstraint(ptr.To[string](string(VAPDefaultYes))),
-			isAdmissionRequest: true,
```
ritazh, line 305: do we not need to test constraint override cases?

ritazh:
Seeing this failure consistently in all the PRs:
> unable to start control plane itself: failed to start the controlplane. retried 5 times: fork/exec /app/.tmp/bin/k8s/1.28.7-linux-amd64/etcd: exec format error

constraint/pkg/client/drivers/k8scel/driver_test.go
JaydipGabani: @ritazh we decided on inheriting behavior from template for constraints. If template has `generateVAP` set to true, we generate VAPB for constraints. And since we are treating VAP as enforcement point we would have it enabled by default and users can ommit VAP with `scopedEnforcementActions` spec field on constraints. But generation property is now inherited from template.

constraint/pkg/client/drivers/k8scel/driver_test.go
JaydipGabani: Constraint is inheriting the generation behavior now and not overriding.

ritazh: But sounds like `scopedEnforcementActions` spec field on constraints can override? if so, then we should include this in the test?

constraint/pkg/client/drivers/k8scel/driver_test.go
JaydipGabani: My bad, should have been clear. With scopedEnforcementActions, user can decide which engine to use, CEL or Rego, for enforcing policies. However generation property is always controlled by `generateVAP` field. Once VAP and VAPB exists in the cluster, intree enforcement will always be there. 

Giving the design a 2nd thought I think this would be a better approach - 

1. `generateVAP` is set to true in CT, generate VAP.
2. `generateVAP` is set to true in CT, generate VAPB for C.
3. `generateVAP` is set to true in CT, VAP enforcementPoint is ommited

constraint/pkg/client/drivers/k8scel/driver_test.go
JaydipGabani: If this above comment makes sense - https://github.com/open-policy-agent/frameworks/pull/427#discussion_r1646515841 - then yes we should include this test also.

constraint/pkg/client/drivers/k8scel/driver_test.go
ritazh: > With scopedEnforcementActions, user can decide which engine to use, CEL or Rego

nit: to be clear, ^ this statement is incorrect. `scopedEnforcementActions` helps users define for a given constraint, which enforcementPoint (gatekeeper webhook or vap/intree or gator) and which action (deny, warn) can be used. users choose which engine to with the following in the constraint template:

```
targets:
    - target: admission.k8s.gatekeeper.sh
      code:
```

constraint/pkg/client/drivers/k8scel/driver_test.go
JaydipGabani: @ritazh since `scopedEnforcementActions` spec doesn't exists in this PR, would it be best if these tests are included again as part of - #403 - where we are adding the spec to be able to override.

constraint/pkg/client/drivers/k8scel/driver_test.go
maxsmythe: Also, bindings would not be generated/not generated by CT.

The logic would be something like:

- If `generateVAP` is set to true in the CT, generate VAP
- If VAP enforcement point is omitted on a constraint, do not generate a VAP Binding
- If VAP enforcement point is enabled on a constraint, generate a VAP binding ONLY if `generateVAP` is true on the CT (arguably we could generate VAP Bindings regardless and leave it to the user to make sure the template is configured appropriately, this may be good for failing closed on a misconfiguration -- thoughts?)

Where a constraint is evaluated is gated purely by the enforcement point. We should not try to infer. Simply: if the user says enforce via Gatekeeper, do that. If the user says enforce via VAP, do that.

In general, we should not try to paper over misconfigurations by trying to infer intent, doing so leads to spaghetti code and weird corner cases.

I realize this is a different approach than before (particularly if we ALWAYS generate a binding regardless of the contents of the template), but multi-enforcement-point lets us avoid the hacky "try to infer things through annotations" we were forced to use before.

ritazh: I always thought it’s odd that k8s allows creations of VAPB without VAP. do you recall why? If we allow VAPB to be created regardless of CT misconfiguration, we miss an opportunity to help users to troubleshoot.  If VAPB got created correctly then it might be harder for the operator to know something is wrong with the CT.

maxsmythe: It may be a principled stance against imperative-type configuration?

>  If we allow VAPB to be created regardless of CT misconfiguration, we miss an opportunity to help users to troubleshoot.

Not necessarily. In fact, with the old behavior we were actually masking potentially undesired state... we used whichever enforcement point made sense given the combination of annotations. A user had no way of saying "use VAP or bust", which could have led to unintentional webhook usage.

Without running inference, we are giving users a bit more power by affirmatively saying "I want VAP". It does come at the cost of now the user needs to be aware of whether VAP is enabled, but Gatekeeper couldn't tell if a VAP definition was successfully installed anyway, since they don't have `status` (it may be that there were compilation errors). The user will also have more control over the correct approach for "enforcement gaps" (as defined by the [user flow doc](https://docs.google.com/document/d/1UELB0dmt10K5jBwETUnSKKQN7qeivOjvi-AvzFTOsHw/edit#heading=h.qr0os9voygyv).

Without the inference logic, what G8r will do is much more simply explained, which should be both easier to write and easier to debug. Eventually users will even be able to use a "fail"-type failure policy to make misconfigurations even more visible.

WRT simplifying debugging...

we can always report a "missing definition" state on the status of the constraint.

We could think about something like webhook validation, but that also has faults. It will lead to transient rejections if users kubectl apply a bulk manifest. It also wouldn't address situations where the definition is removed out from under the constraint.

Perhaps the biggest reason for separating the two is coherence. Either webhook and VAP are separate enforcement points or they are not. I think we already had this discussion and said that they were separate? There was an option for a three-EP approach where we had VAP, VWH and "whichever works", I don't think we opted for that last one, though.

ritazh: I'm ok with always use enforcementAction to determine VAPB generation or not, instead of deciding for users with what is provided in the constraint template. 

Default enforcementAction would include VAP therefore always generate VAPB. If user explicitly omitted VAP from enforcementPoint in constraint, then do not generate VAPB.

constraint/pkg/client/drivers/k8scel/driver.go
```diff
@@ -193,11 +194,9 @@ func (d *Driver) Query(ctx context.Context, target string, constraints []*unstru
...
-
-		// if we assume VAP enforcement for a given constraint/template combo, Gatekeeper
+		// if we assume VAP enforcement for a given template, Gatekeeper
 		// should not be evaluating that constraint/template in an admission context.
-		if isAdmission && assumeVAPEnforcementNotDisabled && wrappedValidator.assumeVAPEnforcement {
+		if isAdmission && wrappedValidator.assumeVAPEnforcement {
```
maxsmythe, line 199: @JaydipGabani I think this goes away with the multi-enforcement action change?

Basically, we always enforce at the webhook level unless the constraint is configured not to.

JaydipGabani: Good catch! missed this one.

constraint/pkg/client/drivers/k8scel/driver_test.go
```diff
@@ -116,7 +94,7 @@ func TestValidation(t *testing.T) {
 		name               string
 		template           *templates.ConstraintTemplate
 		constraint         *unstructured.Unstructured
-		vapDefault         *vapDefault
+		vapDefault         bool
```
ritazh, line 97: you can remove the ones not in use anymore

ritazh (approved):
LGTM

maxsmythe (approved):
LGTM

PR #432, opened 2024-05-16
fix: Create dynamic CEL object variable

maxsmythe:
Defines a variable, `anyObject` that behaves in the same way `request.Object` does in Rego, allowing for a 1:1 behavioral analogue for Rego's `request.Object` command when running a template in VAP.

This also fixes a bug in generated CEL code where user-defined variables are placed before `variables.params` is defined, making users unable to reference `variables.params` in the variables they create.


constraint/pkg/client/drivers/k8scel/transform/cel_snippets.go
```diff
@@ -169,6 +169,23 @@ func BindParamsCEL() []cel.NamedExpressionAccessor {
 	}
 }
 
+func BindObjectV1Beta1() admissionregistrationv1beta1.Variable {
+	return admissionregistrationv1beta1.Variable{
+		Name:       schema.ObjectName,
+		Expression: `has(request.operation) && request.operation == "DELETE" && object == null ? oldObject : object`,
```
ritazh, line 175: This logic for rego is actually in GK not framework: https://github.com/open-policy-agent/gatekeeper/blob/f81f0acacb00802108abc81ba5d862bec6c2d633/pkg/util/request_validation.go#L29, should it be moved to framework to be consistent? not a PR blocker.

constraint/pkg/client/drivers/k8scel/schema/schema.go
```diff
@@ -22,6 +22,8 @@ const (
 	ReservedPrefix = "gatekeeper_internal_"
 	// ParamsName is the VAP variable constraint parameters will be bound to.
 	ParamsName = "params"
+	// ObjectName is the VAP variable that describes either an object or (on DELETE requests) oldObject
+	ObjectName = "anyObject"
```
ritazh, line 26: With this change, do we plan to update all cel expressions for gatekeeper-libraries policies to use this before merging those PRs? if so, then we need to cut a patch release for GK and update GK-lib CI and docs to ensure the latest framework/GK is used.

sozercan: yea, we'll need to merge this, vendor it in GK, cut a GK release (3.16.1), and update library

sozercan (approved):
LGTM pending @ritazh's comments

constraint/pkg/client/drivers/k8scel/transform/cel_snippets.go
maxsmythe: Framework should not have any coupling with any particular platform.

We should probably move the logic to the TargetHandler.

In addition, we should update the engine logic to __undo__ the setting of obj == oldObject on DELETE, otherwise CEL code will behave differently in on-k8s VAP vs in-g8r VAP.

maxsmythe:
LGTY @ritazh ?

ritazh (approved):
LGTM

Looks like there's a race condition failure in the readiness tracker test https://github.com/open-policy-agent/frameworks/actions/runs/9134818421/job/25121118481?pr=432

constraint/pkg/client/drivers/k8scel/transform/cel_snippets.go
ritazh: > We should probably move the logic to the TargetHandler.

Move the `obj == oldObject on DELETE` logic to the TargetHandler instead of the webhook validation handler right?

maxsmythe: WRT moving it to target handler -- right.

The problem with moving it to the Rego driver is that the Rego driver is more generic than the VAP driver.

The VAP driver is highly coupled with K8s/KRM -- it is strongly typed, assuming an inbound admission review object. Because of this, and because the VAP driver is explicitly intended to support VAP (as opposed to a more generic application of CEL) I'm okay with adding "undoing" logic to the driver (we're not increasing coupling).

For something like the Rego driver, the only real assumption is that the inbound object is JSON-structured.

maxsmythe:
WRT the race condition, I'm hoping @David-Jaeyoon-Lee's PR will address this once we solve the shared-fate issue across goroutines:

https://github.com/open-policy-agent/gatekeeper/pull/3308

PR #435, opened 2024-05-19
Use k8s client lib v0.30.1

tamalsaha:


sozercan:
@tamalsaha thanks for the pr! can you run `make vendor manifests` when you get a chance?

tamalsaha:
@sozercan done. PTAL!

davidgamero:
+1

tamalsaha:
There is a breaking api change between 0.30.0 and 0.30.3 . It is worth exploring if that should be fixed here.

ritazh:
> There is a breaking api change between 0.30.0 and 0.30.3 . It is worth exploring if that should be fixed here. https://github.com/kubernetes/kubernetes/blob/6834a1ef3d928f6f7e00e801a43c37bead75dbe9/pkg/apis/admissionregistration/validation/validation.go#L1304
> 
> `compiler, err := plugincel.NewCompositedCompiler(environment.MustBaseEnvSet(environment.DefaultCompatibilityVersion(), strictCost))`
> 
> The `strictCost` boolean parameter has been newly added

We could deal with it as a separate PR to keep this PR small. wdyt?

tamalsaha:
> > There is a breaking api change between 0.30.0 and 0.30.3 . It is worth exploring if that should be fixed here. https://github.com/kubernetes/kubernetes/blob/6834a1ef3d928f6f7e00e801a43c37bead75dbe9/pkg/apis/admissionregistration/validation/validation.go#L1304
> > `compiler, err := plugincel.NewCompositedCompiler(environment.MustBaseEnvSet(environment.DefaultCompatibilityVersion(), strictCost))`
> > The `strictCost` boolean parameter has been newly added
> 
> We could deal with it as a separate follow up PR to keep this PR smaller. wdyt?

Sounds good to me.

sozercan:
Looks like this breaks GK tests (these are informative), we'll need to handle that when we vendor this in.

sozercan (approved):
Thank you! LGTM

PR #437, opened 2024-05-24
fix: `anyObject` was not fully plumbed to compiler, streamline plumbing to fix

maxsmythe:
variables.anyObject isn't recognized in-process until this is merged.


sozercan (approved):
LGTM

PR #449, opened 2024-07-03
fix: Unset Object on DELETE requests when using the VAP driver

maxsmythe:
This is necessary so that in-Gatekeeper VAP code evaluation behaves identically to in-K8s VAP code evaluation.


JaydipGabani (approved):

ritazh (approved):
LGTM

PR #479, opened 2024-09-26
chore: renaming gatekeeper ci test to add informatinve

JaydipGabani:
Informative tests are expected to fail for changes such as API change or OPA version bump or other changes that may not be vendored in on Gatekeeper side yet.


.github/workflows/gatekeeper.yml
```diff
@@ -6,7 +6,7 @@ permissions:
 
 jobs:
   gk-test:
-    name: "Gatekeeper Test"
+    name: "[Informative] Gatekeeper Test"
```
ritazh, line 9: should we add `continue-on-error: true`

JaydipGabani: I would think we still want it to fail and see what is causing the error. If it is changes that are limited to frameworks and test fails, then we may not want to merge those changes. For example, refactoring in frameworks should not break GK test.

ritazh: you are right. I didn't know that it would show the gatekeeper_test job as succeed.

ritazh (approved):

PR #517, opened 2025-01-23
constraint: Update Rego driver to support rego.v1

charlieegan3:
Updated to follow https://github.com/open-policy-agent/frameworks/pull/518 with support for rego.v1 in the rego driver.

The obvious thing to consider here is that attempting parsing in more than one rego version is not ideal. This is done here since we will not know the source version of the rego until we try to parse it. We might consider something like this to allow users to force GK to use a given version.

```yaml
apiVersion: templates.gatekeeper.sh/v1beta1
kind: ConstraintTemplate
metadata:
  name: k8srequiredlabels
spec:
  crd:
    spec:
      names:
        kind: K8sRequiredLabels
      validation:
        # Schema for the `parameters` field
        openAPIV3Schema:
          properties:
            labels:
              type: array
              items:
                type: string
  targets:
    - target: admission.k8s.gatekeeper.sh
      version: v1 <------ this, or something like it.
      rego: |
        package k8srequiredlabels

        violation contains {"msg": msg, "details": {"foo": "bar"}} if {
          ...
        }
```

### DIFF
--- a/constraint/pkg/client/drivers/rego/compilers.go
+++ b/constraint/pkg/client/drivers/rego/compilers.go
@@ -163,11 +163,10 @@ func parseConstraintTemplateTarget(rr *regorewriter.RegoRewriter, targetSpec *te
 
 	var version ast.RegoVersion
 	switch regoSrc.Version {
-	case "v0":
-		version = ast.RegoV0
 	case "v1":
 		version = ast.RegoV1
 	default:
+		// v0 and any other value is v0
 		version = ast.RegoV0
 	}
 


maxsmythe:
Thanks for the PR!

To answer your question about regorewriter:

First some background:

regorewriter was written when all constraint templates were compiled into the same OPA environment. It was intended to provide hermeticity. Basically, we wanted to be sure that no constraint template could modify the behavior of any other constraint template, so we rewrote the module paths to ensure non-overlap. We also limited the sub-fields of `data` that could be accessed in order to prevent cross-querying. This has the side effect of providing static verification to users (e.g. catching mis-types like `data.inventry` instead of `data.inventory`.

Since that time, we have moved on to compiling each constraint template into a separate Rego environment (though the data cache is shared). This was mainly to reduce the O(n^2) complexity growth of compiling templates, but had the effect of strengthening sandboxing as well.

As such, regorewriter is less load bearing than it used to be, but still useful.

Key features of regorewriter (I may be missing some, did a brief survey of the code):

- ensure all library modules have the expected prefix (`data.libs`) so they can be imported
- ensure that only recognized subfields of `data` are used -- this helps avoid typos, but more importantly it gives us room to add future subfields without breaking backwards compatibility
- make sure `data` cannot be reassigned to a new variable to escape static analysis (`data` must always have a sub-reference)
- do not let any package clobber a root field (e.g. `data.libs` is not valid, `data.libs.toolbox` is)

It does look like some portions of regorewriter may be redundant -- if we require library packages to start with `data.libs`, why are we also rewriting those references? But a close reading of the code would be required to determine that for sure.

WRT adding the version string...

I like it! Though we should probably nudge people to using the newer format, which lives under `spec.targets[].engine.source`, example:

https://github.com/open-policy-agent/gatekeeper-library/blob/99a8a79972adcaf4c521e5d3ba87c5a6a1643049/library/pod-security-policy/forbidden-sysctls/template.yaml#L78-L150

I'd make it a sibling of the `rego` field, defined here:

https://github.com/open-policy-agent/frameworks/blob/df595163a8ba4bfcd4d4337b4249d6bf2f5460fe/constraint/pkg/client/drivers/rego/schema/schema.go#L19-L24

for backwards compatibility, undefined -> v0, and using `spec.targets[].rego` would imply v0

charlieegan3:
Hey, thanks for the input here Max and sorry for the delay in replying. Now we have https://github.com/open-policy-agent/frameworks/pull/518 in I think we are good to continue here.

1. Do you think we should add the version field as part of this PR? (happy to work on that here but also ok to make this smaller chunks if preferred)
2. Thanks for the helpful summary re: regorewriter. Is there more you'd like to see here? My feeling is that the current changes keep it up to date but I'm not sure we need to address bugs like the `some x in data.lib` issue here. Do you agree?

charlieegan3:
Thanks for the input @JaydipGabani 👍

JaydipGabani:
@charlieegan3 I was trying to test this change with v0 and v1 syntax. When CT had v0 syntax everything worked as expected -
```rego
violation[{"msg": msg, "details": {"missing_labels": missing}}] {
        provided := {label | input.review.object.metadata.labels[label]}
        required := {label | label := input.parameters.labels[_]}
        missing := required - provided
        count(missing) > 0
        msg := sprintf("you must provide labels: %v", [missing])
      }

```


However when I switched to using v1 syntax as below - 

```rego
violation contains [{"msg": msg, "details": {"missing_labels": missing}}] if {
        provided := {label | some label; label = input.review.object.metadata.labels[label]}
        required := {req_label | some req_label; req_label = input.parameters.labels[_]}
        missing := required - provided
        count(missing) > 0
        msg := sprintf("you must provide labels: %v", [missing])
      }
``` 

I started having the error - 
```
"Could not ingest Rego: unable to compile modules: 2 errors occurred:\nhooks.hooks_builtin:23:
          rego_type_error: object.get: invalid argument(s)\n\thave: (array<object<details:
          object<missing_labels: any<number, set[any]>>, msg: string>>, string, object[any:
          any], ???)\n\twant: (object: object[any: any], key: any, default: any, value:
          any)\nhooks.hooks_builtin:24: rego_type_error: undefined ref: r.msg\n\tr.msg\n\t
          \ ^\n\t  have: \"msg\"\n\t  want (type): number"
```

I couldn't figure out how to fix v1 syntax, could you verify if I am using the correct syntax for v1?

charlieegan3:
Hey, thanks for the feedback here. I'd missed some tests for the add template operation. I have added some in 478d97c5bcc40a7c0440621a37452b2cc13b4f3b which is based on the bench rego examples. Lmk if that looks good, I think that should work now.

JaydipGabani:
@charlieegan3 Thanks for adding tests.

JaydipGabani (approved):
lgtm

charlieegan3:
Thanks @JaydipGabani! 

Re:

> Do you think we should add the version field as part of this PR?

Any more thoughts on this point? I'm happy to do it here or in a follow on PR, but just keen to stick to whatever you all prefer.

JaydipGabani:
@maxsmythe @ritazh @sozercan bump on this^^

constraint/pkg/regorewriter/regorewriter.go
```diff
@@ -124,7 +124,17 @@ func (r *RegoRewriter) addFileFromFs(path string, slice *[]*Module) error {
 		return fmt.Errorf("%w: %v", ErrReadingFile, err)
 	}
 
-	m, err := ast.ParseModule(path, string(bytes))
+	attemptVersions := []ast.RegoVersion{ast.RegoV0, ast.RegoV1}
```
ritazh, line 127: why is this using v0 first then v1 while https://github.com/open-policy-agent/frameworks/pull/517/files#diff-f21596183dbd319da5a0fe99596c9a346cc60a7222b062b6beb5342a74242724R242 in compilers.go parses to v1 first then v0? Was this intentional? If so, maybe add a comment to explain the rationale?

Since no version in CT means v0, does it make sense to default to v1 always and fallback to v0?

charlieegan3: Good spot, I'd missed updating this to be the same order as parsing in 947c51b957eddd426fd0cc27f4cbcf67751ef26b.

> Since no version in CT means v0

I was under the impression we would add the version field in another PR. But I can add it in this PR instead if that's preferred. I think if we have versioned CRs, then we might as well drop this multi-attempt parsing functionality. This was intended just to be an interim option to allow folk to use v1 Rego before we had controllable versioning. Lmk what you prefer and I can do that.

ritazh: Based on past discussions, we want users to opt-in to rego v1 by explicitly adding:
```yaml
  targets:
    - target: admission.k8s.gatekeeper.sh
      code:
      - engine: Rego
        source:
          version: v1
          rego: |
```
When they don't provide version, then we assume v0 to ensure backward compatibility.

Given this, should we default to parse v0 first, and if failed, then parse with v1?

charlieegan3: Ahh, ok I can get that done in this PR instead. (https://github.com/open-policy-agent/frameworks/pull/517#issuecomment-2680124873)

In which case, I will remove the attemptVersions parsing and only parse v0 unless the source.version field is set to v1. 

I can get that done this week.

charlieegan3: Hey, I have just pushed 37bcfcbe6c038382d4ab1f3971cd99ee121579bd which does this.

JaydipGabani:
@charlieegan3 seems like unti tests are failing. PTAL.

charlieegan3:
Sorry about that, there were some higher level tests I'd missed. I have patched it up now.

constraint/pkg/client/drivers/rego/rego.go
```diff
@@ -51,7 +50,7 @@ var hookModule *ast.Module
 
 func init() {
 	var err error
-	hookModule, err = parseModule(hookModulePath, hookModuleRego)
+	hookModule, err = parseModule(hookModulePath, ast.RegoV1, hookModuleRego)
```
ritazh, line 53: why is the version set to regov1 here?

constraint/pkg/client/drivers/rego/rego.go
```diff
@@ -18,31 +18,30 @@ const (
...
-  data.template.violation[r] with input as inp
-
-  # Construct the response, defaulting "details" to empty object if it is not
-  # specified.
-  response := {
-    "key": key,
-    "details": object.get(r, "details", {}),
-    "msg": r.msg,
-  }
+violation contains response if {
```
ritazh, line 24: same, this should be v0

constraint/pkg/client/drivers/rego/schema/schema.go
```diff
@@ -49,22 +52,36 @@ func GetSource(code templates.Code) (*Source, error) {
...
 	}
+
 	source.Rego = rego
 
+	version, found, err := unstructured.NestedString(v, "version")
+	if err != nil {
+		return nil, fmt.Errorf("%w: while extracting Rego version", err)
+	}
+	if !found {
+		version = "v0"
```
ritazh, line 73: is there a test that validates this

constraint/pkg/client/drivers/rego/rego.go
charlieegan3: Hey, this is because the Rego syntax for the hookModuleRego is v1 now. It's possible for OPA to work with more than one version of Rego at the same time. E.g.

```
…/opa/example main ➜ cat v1/.manifest

       │ File: v1/.manifest

   1   │ {
   2   │   "revision" : "7864d60dd78d748dbce54b569e939f5b0dc07486",
   3   │   "rego_version": 1,
   4   │   "file_rego_versions": {
   5   │     "/foo.rego": 0,
   6   │     "/bar.rego": 1
   7   │   }
   8   │ }

…/opa/example main ➜ cat v1/foo.rego

       │ File: v1/foo.rego

   1   │ package foo
   2   │
   3   │ messages[msg] := true {
   4   │     msg := "foo"
   5   │ }

…/opa/example main ➜ cat v1/bar.rego

       │ File: v1/bar.rego

   1   │ package bar
   2   │
   3   │ allow if input.user == "admin"

…/opa/example main ➜ opa eval -b v1 data
{
  "result": [
    {
      "expressions": [
        {
          "value": {
            "bar": {},
            "foo": {
              "messages": {
                "foo": true
              }
            }
          },
          "text": "data",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}

```

ritazh: yea, I think the question is should we convert hookModuleRego to v1 now? what are the pros and cons with that? since opa supports both versions at the same time, should we slowly migrate things from v0 to v1 while supporting users who want to opt-in to v1?

constraint/pkg/client/drivers/rego/compilers.go
```diff
@@ -159,7 +160,18 @@ func parseConstraintTemplateTarget(rr *regorewriter.RegoRewriter, targetSpec *te
 	if err != nil {
 		return nil, err
 	}
-	entryPoint, err := parseModule(templatePath, regoSrc.Rego)
+
+	var version ast.RegoVersion
+	switch regoSrc.Version {
+	case "v0":
```
JaydipGabani, line 166:
```suggestion
	case "v1":
		version = ast.RegoV1
	default:
		version = ast.RegoV0
```

constraint/pkg/client/clienttest/cts/opts.go
```diff
@@ -1,30 +1,33 @@
...
   msg := "denied"
 }
 `
-	MockTemplateName  string = "fakes"
-	MockTemplate      string = "Fakes"
-	MockTargetHandler string = "foo"
+	MockTemplateName  string          = "fakes"
+	MockTemplate      string          = "Fakes"
+	MockTargetHandler string          = "foo"
+	RegoVersion       ast.RegoVersion = ast.RegoV0
```
JaydipGabani, line 24: Where is `RegoVersion` being used?

JaydipGabani (changes requested):

constraint/pkg/client/drivers/rego/rego.go
charlieegan3: I don't think there are any downsides to doing this now, but have reverted to v0 in f758a482d8836fc6a3b532c38d87d79fd0680f19
 to keep the changeset small so we can move this faster.

constraint/pkg/client/drivers/rego/schema/schema.go
charlieegan3: Thanks for flagging, one added in fee23bf54352a13758430f9ebf22a7f68c7c33fd

constraint/pkg/client/drivers/rego/compilers.go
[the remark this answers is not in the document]
charlieegan3: Thanks, Fixed in 0c1d4af0a3aa40fbffa5d2e3207ae0b3129658ca

constraint/pkg/regorewriter/regorewriter.go
```diff
@@ -165,24 +167,40 @@ func (r *RegoRewriter) addPathFromFs(path string, slice *[]*Module) error {
...
-	return r.addPathFromFs(path, &r.entryPoints)
+	// TODO: use v0 for regorewriter until version can be determined
+	return r.addPathFromFs(path, ast.RegoV0, &r.entryPoints)
+}
+
+// AddBaseFromFsV1 adds a base source which will not have it's package path
+// rewritten.  These correspond to the rego that will be populated into a
+// ConstraintTemplate with the 'violation' rule.
+// The Rego path is parsed as RegoV1.
+func (r *RegoRewriter) AddBaseFromFsV1(path string) error {
```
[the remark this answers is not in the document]
charlieegan3: I've removed these additional functions now in 4b17f27bb7730719726920ec9a8cc66ae3bc6572. I had hoped to set the direct for use from rego-rewriter but that can perhaps come in a later PR.

constraint/pkg/regorewriter/regorewriter.go
```diff
@@ -165,24 +167,40 @@ func (r *RegoRewriter) addPathFromFs(path string, slice *[]*Module) error {
...
 // AddLibFromFs adds a library source which will have the package path updated.
 func (r *RegoRewriter) AddLibFromFs(path string) error {
-	return r.addPathFromFs(path, &r.libs)
+	// TODO: use v0 for regorewriter until version can be determined
+	return r.addPathFromFs(path, ast.RegoV0, &r.libs)
+}
+
+// AddLibFromFsV1 adds a library source which will have the package path updated.
+// The Rego path is parsed as RegoV1.
+func (r *RegoRewriter) AddLibFromFsV1(path string) error {
```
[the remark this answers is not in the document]
charlieegan3: See https://github.com/open-policy-agent/frameworks/pull/517#discussion_r2011781796

charlieegan3:
Hey both, thanks for the review here. Had a long weekend this side but all patched up now 🙏

ritazh (approved):
LGTM

constraint/pkg/regorewriter/regorewriter.go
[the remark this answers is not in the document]
JaydipGabani: +1, on merging this PR.

JaydipGabani (approved):
lgtm

constraint/pkg/client/clienttest/cts/opts.go
JaydipGabani: I think this may have been missed. I am ok with merging this PR. Or I can raise a follow up to remove this asap. thoughts?

ritazh: hmm i think the intent might be to use it as part of the defaults https://github.com/open-policy-agent/frameworks/pull/517/files/dc39c8d9ba4422d4229d5d72fec977bb715e8d59#diff-cbc675a4ccf1d3ea615bc28ddc01ee192bcf11a9a4a206caa462cdac8454317aR30

charlieegan3:
Thanks both for the reviews here and sorry for the delay in getting it patched up this week.

Open to input on next step ideas:

- update the internal Rego to be v1
- better understand (for me!) the requirements of the rego rewriter with v1 in mind.
- ... likely others.

We can also discuss on https://github.com/open-policy-agent/gatekeeper/issues/3577 if you'd prefer.

PR #560, opened 2025-05-15
chore: bumping go in CI and fixing test errors

JaydipGabani:


.github/workflows/gatekeeper.yml
```diff
@@ -40,5 +40,5 @@ jobs:
         working-directory: go/src/github.com/open-policy-agent/gatekeeper
 
       - name: Make test
-        run: make test
+        run: make native-test
```
ritazh, line 43: Whats the rationale behind changing this?
https://github.com/open-policy-agent/gatekeeper/blob/master/Makefile#L191-L193 calls `gatekeeper-test make native-test`

JaydipGabani: `make test` runs tests with root user, resulting in failure of the test - [Error walking directory](https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/export/disk/disk_test.go#L829) - where the permissions are set to root to deny the access to the dir. Switching it to `make native-test` makes sure that we do not run tests with root user.

ritazh (approved):
LGTM

.github/workflows/gatekeeper.yml
ritazh: Great catch @JaydipGabani ! For future PRs, if you could include the reasons for the change as a comment next to the code or a PR comment, it could help speed up reviews. thanks again for the PR!

PR #595, opened 2025-08-12
Change SemanticEqual to include annotations for constraint caching

nreisch:
Constraint cache remains stale when annotations are updated. Although the annotations don't impact driver enforcement, the annotations may contain state that other parts of the system need such as in audit logging for export violations, thus it's useful to refresh the cache when annotations are updated.


JaydipGabani (approved):

JaydipGabani:
@nreisch can you fix the lint?

PR #706, opened 2026-04-09
Avoid nil-pointer panic when Rego template source is missing

sozercan:
### Motivation
- `GetSource` in the Rego schema driver dereferenced `code.Source` without checking for `nil`, allowing a malformed ConstraintTemplate with `engine: Rego` and missing/`null` `source` to cause a nil-pointer panic and crash the process.

### Description
- Add a nil guard in `constraint/pkg/client/drivers/rego/schema.GetSource` to return `ErrMissingField` when `code.Source` is `nil` or `code.Source.Value` is `nil`, preventing dereference of a nil pointer.
- Add unit tests in `constraint/pkg/client/drivers/rego/schema/schema_test.go` to cover both `Source: nil` and `Source` with an empty `Anything` value and assert `ErrMissingField` is returned.
- Preserve existing behavior for valid `source` payloads and error semantics for bad types or missing required `rego` field.

### Testing
- Ran `go test ./pkg/client/drivers/rego/schema` from the `constraint/` module and the package tests passed successfully.
- Attempting to run `go test` from the repository root path that is not a Go module produced a module initialization error, so tests were executed from the `constraint` module instead.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d7e5d5727c832a92c438bc50e4b0cf)


JaydipGabani (approved):
lgtm

JaydipGabani:
@dependabot rebase

